PR #52, opened 2019-01-23
Fixed issue where modal open and close methods where throwing console errors when used with out a modal trigger button

levimcg:
This PR should fix console errors when opening and closing the modal programmatically without the presence of a button/element with a `data-modal-trigger` attribute used to open the modal.

This was reported to me directly offline, so no issue currently exists. The errors do not break the functionality of the modal opening or closing, but to prevent the `customEvent`s from firing off.

All the modal tests are passing, and I tested locally opening and closing the modal without the presence of a trigger element i.e. using only the `.open()` and `.close()` methods.

After a review, I think it would be good to push this out as a minor/patch release.

### DIFF
--- a/src/js/components/modal.js
+++ b/src/js/components/modal.js
@@ -246,8 +246,9 @@ var Modal = (function() {
         event.preventDefault();
 
         close(id);
-
-        activeTrigger.focus();
+        
+        // Check to make sure modal was opened via a trigger element.
+        if (activeTrigger !== null) activeTrigger.focus();
 
         break;
       case trigger === id && !event.clickedInModal:
--- a/src/js/components/modal.js
+++ b/src/js/components/modal.js
@@ -119,11 +119,11 @@ var Modal = (function() {
      * be deprecated in the next major version.
      */
     if (typeof id === 'object' && id.nodeType === 1) {
-        id = id.getAttribute('id');
+      id = id.getAttribute('id');
 
-        if (!id) {
-          throw new Error('Please proved an id attribute for the modal you want to close.');
-        }
+      if (!id) {
+        throw new Error('Please proved an id attribute for the modal you want to close.');
+      }
     }
 
     var modal = _createModalObject(id);
@@ -235,7 +235,7 @@ var Modal = (function() {
       (trigger.getAttribute(CLOSE_ATTR) && trigger.getAttribute(CLOSE_ATTR) !== 'close' ? trigger.getAttribute(CLOSE_ATTR) : false) ||
       event.target.closest(MODAL_SELECTOR);
 
-    switch (trigger !== null || undefined) {
+    switch (trigger !== null) {
       case trigger.hasAttribute(TRIGGER_ATTR):
         open(id);
 
@@ -258,6 +258,8 @@ var Modal = (function() {
         close(id);
 
         activeTrigger.focus();
+        
+        break;
       default:
         return;
     }
--- a/src/js/components/modal.js
+++ b/src/js/components/modal.js
@@ -82,8 +82,25 @@ var Modal = (function() {
 
     // Sets a class on the body to handle overflow and scroll.
     document.body.classList.add('rvt-modal-open');
-
-    fireCustomEvent(modal.trigger, TRIGGER_ATTR, 'modalOpen');
+    
+    /**
+     * TODO: Here we are checking to see if the modal was triggered by
+     * clicking an element with the "data-modal-trigger" attribute. This
+     * is to cover the case that someone might use a modal without a
+     * trigger associated with it i.e. opening the modal programmatically
+     * via the API. This could probably be refactored to always use the
+     * modal's "id" attribute when emitting the custom event. Same for the
+     * ".close()" method below.
+     */
+    
+    // If there is a corresponding modal trigger
+    if (activeTrigger !== null) {
+      // Fire the custom event from the trigger
+      fireCustomEvent(activeTrigger, TRIGGER_ATTR, 'modalOpen');
+    } else {
+      // Otherwise use the modal's id attribute
+      fireCustomEvent(activeModal, 'id', 'modalOpen');
+    }
 
     if (callback && typeof callback === 'function') {
       callback();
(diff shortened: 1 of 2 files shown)


src/js/components/modal.js
```diff
@@ -82,8 +82,25 @@ var Modal = (function() {
...
+     * trigger associated with it i.e. opening the modal programmatically
+     * via the API. This could probably be refactored to always use the
+     * modal's "id" attribute when emitting the custom event. Same for the
+     * ".close()" method below.
+     */
+    
+    // If there is a corresponding modal trigger
+    if (activeTrigger !== null) {
+      // Fire the custom event from the trigger
+      fireCustomEvent(activeTrigger, TRIGGER_ATTR, 'modalOpen');
```
basham: This seems like an odd choice. The modal is the thing that is opening or closing, so the event should only dispatch from the modal, not the trigger. That would simplify what Rivet users have to do. They don't have to think about if they have a trigger or not and where the event is bubbling from. It should be from the one, guaranteed place (the modal).

basham:
I like this change. That allows more custom things to happen, such as manually changing the focus upon modal close, rather than relying on whatever the trigger was.

```js
document.addEventListener('modalClosed', (event) => {
  if (event.detail.name() === 'my-modal')
    window.requestAnimationFrame(() => document.getElementById('custom-trigger').focus())
})
```

src/js/components/modal.js
levimcg: Right, that's why I put that comment explaining it above. I'm hesitant to totally change the functionality now, but I guess there really would never be a time when there wouldn't be an `id` attribute present on the modal itself. I just wouldn't work. It just makes me nervous 😬

basham: If there's no `id` on the modal, then throw an error.

levimcg: We're already doing that in the `.open()` method, so I guess it's covered in that sense.

https://github.com/indiana-university/rivet-source/blob/develop/src/js/components/modal.js#L66

basham:
So, right now, the modal looks for a particular trigger via the data attribute, so it can return focus to it upon modal close. But here are some things I'd like the modal to cover:
1. Before opening the modal, we know the trigger that should get focus upon modal close.
2. We only know the trigger that should get focus until after the modal closes. For example, maybe the modal is submitting something via a server request, and it behaves differently depending on success or error.
3. We don't care to manage or know about the particular trigger. Just return focus to the item last in focus before the modal opened.

Can all these things be covered with the current API?
https://rivet.iu.edu/components/overlays/modals/

src/js/components/modal.js
```diff
@@ -335,9 +311,9 @@ var Modal = (function() {
         if (activeModal.hasAttribute('data-modal-dialog')) return;
 
         close(activeModal.id);
-
-        activeTrigger.focus();
-
+
```
basham: Remove unused indentation.

src/js/components/modal.js
```diff
@@ -335,9 +311,9 @@ var Modal = (function() {
         if (activeModal.hasAttribute('data-modal-dialog')) return;
 
         close(activeModal.id);
-
-        activeTrigger.focus();
-
+        
+        if (activeTrigger !== null) activeTrigger.focus();
+
```
basham: Remove unused indentation.

levimcg: Sorry, not sure which indentation you're talking about here. Everything is indented one tab stop (2 spaces) in from the `case KEYS.ecape:` switch statement above. How do you expect to see it indented instead?

basham: There are 8 spaces (`        `) on line 314 and 316, even though there is no code on those lines. I guess I leave that as a code style that you should manage, but it seems odd to have indentation on lines without code.

levimcg: Aaaah, ok, yeah, I see the spaces now. That reminds me that we do have a JS lint task set up that I should re-enable. I had removed when I was converting everything over to two spaces, and forgot to add it back. I'll do that as a separate PR and fix up any style errors it might be throwing. Probably would have caught those extra spaces.

levimcg:
@basham RE: your questions above about he API. Theoretically, I _think_ these thing should be possible via the [current API using the optional callbacks](https://rivet.iu.edu/components/overlays/modals/#javascript-api) in the `.open()` and `.close()` methods. I'd be interested in seeing your use case as an example.

I've been thinking a lot about future API changes where we might move to a design where each of Rivet's JS components are constructors, i.e. need to be explicitly instantiated. So something like:

```js
const myModal = new RivetModal(options);
```

Where `options` would be an object you could pass in explicitly setting potentially all or some of the the things you mention above.

I think that would make it much easier to provide the kind of fine-grained control we're talking about here. The goal originally was to just make everything _plug-and-play_—include the script on the page and everything "Just Works™". I think that covers _most_ of the typical use cases, but I can see a lot of value in each modal, for example, being its own instance, especially for folks using the Rivet JS alongside libraries, like React, etc.

basham:
Glad you've been pondering the APIs. I've been thinking about them as well.

When thinking about how APIs work currently in the browser, it seems like the API is an extension to the HTML node itself. Input example:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

```js
inputRef.value
inputRef.click()
inputRef.setSelectionRange()
```

When creating web components, you extend `HTMLElement` and the API is exposed on the node itself.

So, what if we do something similar with Rivet?

```js
// Use a CSS selector to convert a single node or a bunch of nodes to a Rivet Modal.
window.rvt.Modal(selector, options)
const modal = document.getElementById('my-rivet-modal')
modal.open()
modal.close()
```

Maybe there could be some way to use `window.rvt.Modal` as a foundational piece, so that you could do this mounting more automatically with [`wickedElements`](https://github.com/WebReflection/wicked-elements) (as a Rivet Add-on).

```
// All nodes that use the .rvt-modal class could automatically be converted to a Rivet Modal.
wickedElements.define('.rvt-modal', ...)

<div class="rvt-modal">...</div>
```

levimcg:
Yeah, I was looking at the regular/wickedElements stuff the other day and it seem super interesting to me. I'd like to spend some time playing around with them for sure. I'm all for making the APIs match native APIs more closely 👍

Issue #24, opened 2019-01-17
add "none" alias to "-p-remove" and "-m-remove" classes


scottanthonymurray:
Several developers (myself included) have expected for there to be a `-none` version of the `rvt-m-[side]-[size]` classes, e.g. `rvt-m-bottom-none`. While `rvt-m-bottom-remove` does it exist, it has proved to be unintuitive based on the naming of the other classes.

I propose adding a `*-none` alias to any class that includes `*-remove`, in order to better meet developer expectations.

Additionally, documentation on the spacing page (https://rivet.iu.edu/components/layout/spacing/#no-spacing) should be updated to include the alias, as well as to spell out the different `-none` classes that are available. It currently only documents `rvt-m-all-remove` and `rvt-p-all-remove`, without mentioning that similar classes exist for top, right, bottom, and left as well.

levimcg:
Closed via #55

Issue #31, opened 2019-01-17
Make styles to remove padding/margin responsive


levimcg:
We have styles to remove padding/margin, but we need them to be responsive.  This will allow us to set padding and margin at a lower resolution and then remove it for higher resolutions. For example, in a form, the fields may wrap to a new line at a lower resolution, so we may need to add some padding between fields, but then when the fields return to their previous position at a higher resolution, the padding is no longer needed.

I have an example of this at the following url:

https://codepen.io/krishancock/pen/NaqWWp

I've written css to remove padding-bottom for medium screens (740px) and up.

Please let me know if you need any additional information.

levimcg:
Support for response `-remove`/`-none` spacing utility classes was (mostly) added via #55 

I realized that I forgot to add the `-none` utilities for the `-m-tb-`, `-m-lr-`, `-p-tb-`, `-p-lr-` spacing classes. I can address in a separate PR.

levimcg:
Closed via #63 and #24

PR #55, opened 2019-01-29
Added none alias to spacing classes

pacowong99:
Can use `rvt-m-all-none`, `rvt-p-all-none`, or other spacing classes to remove padding or margins. From issue #24. Still need to edit the documentation to reflect this change.


levimcg:
@pacowong99 I pushed some updates back to this branch including some comments about how we've historically used the `rvt-` prefix.

I also added some code that generates responsive versions of the `-none` classes. See #31 

This PR will close #24

Issue #53, opened 2019-01-24
Select Down Key Ignored


38brandonwilliams:
## Bugs, issues, broken stuff

When using a select input with a size attribute in a modal with Rivet 1.2.0 JavaScript, down key events are ignored. Down key events should behave like up key events where the selected option changes as a result of the event.

## Steps to reproduce the problem

1. Open a modal
2. Click a select input option
3. Press the down key

## Context

- Browsers with error: Chrome, Edge, Internet Explorer

## Additional Info

Pen with Rivet 1.0.0 code where down key works: https://codepen.io/bw38/pen/bzVeMz
Pen with Rivet 1.2.0 code where down key is ignored: https://codepen.io/bw38/pen/rPOLrx

The problem code is probably the event.preventDefault() statement in drawer.js as shown below.

function _handleKeydown(event) {
    // Handle keyboard stuff
    switch (event.keyCode) {
      case KEYS.down:
        // Stop the page from shimmying when using down key
        event.preventDefault();
        ...

levimcg:
@38brandonwilliams Thanks for creating this issue! We'll have a look and get a fixed prioritized and in [the backlog](https://github.com/indiana-university/rivet-source/projects/1).

levimcg:
@38brandonwilliams Just a quick follow-up. I think I found what is causing the down arrow key not to work. We noticed another issue where the down arrow key was being ignored when used inside of inputs with the `type="number"` attribute as well. I think it's as a result of this `preventDefault()` statement here:

https://github.com/indiana-university/rivet-source/blob/master/src/js/components/drawer.js#L209

I try to confirm with a test case asap, but I _think_ that's it.

nbannist:
I just noticed this today. It also prevents using the keyboard to scroll down through content so it's a big accessibility concern.

PR #66, opened 2019-02-11
Removed preventDefault, and added input selector example to modal for…

illusivesunrae:
… testing. Handles issue #53.


illusivesunrae:
Attempted several different fixes for the scrolling issue caused by code's removal (to no avail). The scrolling issue seems most apparent in header examples using the drawer.

Issue #54, opened 2019-01-29
Loading indicator button color


thrclark:
Question about colors used in the Rivet loading indicator buttons...the default disabled button coloring is grey, but in the 'loading indicator' state it's blue (even though 'loading' is also a disabled state):

<img width="174" alt="screen shot 2019-01-29 at 12 33 16 pm" src="https://user-images.githubusercontent.com/5652113/51929216-ad07a680-23c5-11e9-8508-341da39b83f9.png">

It's possible that a button's previous color (before the 'loading' styles are applied) could be red, green, or blue. In which case, would it seem more reasonable to use the neutral grey coloring instead of blue?

illusivesunrae:
Merged into master in PR #207.

PR #68, opened 2019-02-12
Loading button status color

thrclark:
PR that addresses the conversation thread here regarding the button loading indicator status: https://github.com/indiana-university/rivet-source/issues/54

This takes the color values from the standard disabled button styling and applies them to the loading button status styling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ cypress/screenshots
 /js
 /index.html
 /rivet.zip
-*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ cypress/screenshots
 /js
 /index.html
 /rivet.zip
+*.iml
--- a/src/sass/components/_buttons.scss
+++ b/src/sass/components/_buttons.scss
@@ -280,16 +280,16 @@
 
 .rvt-button--loading,
 .rvt-button--loading[disabled] {
-  background-color: $color-blue--100;
-  border-color: $color-blue--200;
-  color: $color-blue--600;
+  background-color: $color-black--100;
+  border-color: $color-black--400;
+  color: $color-black--700;
   justify-content: center;
   position: relative;
 
   &:hover {
-    background-color: $color-blue--100;
-    border-color: $color-blue--200;
-    color: $color-blue--600;
+    background-color: $color-black--100;
+    border-color: $color-black--400;
+    color: $color-black--700;
   }
 
   .rvt-button__text {


levimcg:
This looks good, @thrclark Do you think we should leave the spinner the blue color or make it a muted gray sort of like the text of disabled buttons? See screenshots:

image
image

thrclark:
really a toss up for me...although one argument for the grey is that it might transition to & from all states (primary, success, danger) a little more smoothly.

basham:
I would absolutely advocate for coloring the indicator gray (the same as the text color). Anything else will just lead to confusion.

levimcg:
@basham I agree. Seems like carrying through the same gray color to the spinner makes the most sense here. @thrclark good call on the transition.

So you would need to add some rules [to this block](https://github.com/indiana-university/rivet-source/blob/8d9a30104bdc091dbcd3dec5f217ab7b7c3ba7d1/src/sass/components/_buttons.scss#L299) in `_buttons.scss` like:

```css
border-top-color: $color-black--400;
border-right-color: $color-black--400;
border-bottom-color: $color-black--400;
```

basham:
@levimcg If the button is disabled, there shouldn't be any hover effect, correct?

levimcg:
@basham No, I wouldn't expect there to be any hover change other than the cursor changing to `not-allowed` like it does currently on disabled buttons.

.gitignore
```diff
@@ -13,4 +13,5 @@ cypress/screenshots
 /sass
 /js
 /index.html
-/rivet.zip
\ No newline at end of file
+/rivet.zip
```
levimcg: Can you add a line of blank space here with a comment what this type of file is? so:

```gitignore
/rivet.zip

# Editor files, etc.
*.iml
```

src/sass/components/_buttons.scss
```diff
@@ -280,16 +280,16 @@
...
   position: relative;
 
   &:hover {
-    background-color: $color-blue--100;
-    border-color: $color-blue--200;
-    color: $color-blue--600;
+    background-color: $color-black--100;
+    border-color: $color-black--400;
+    color: $color-black--700;
   }
```
levimcg: Would need to add styles to change the border color of the `.rvt-loader` while in the `--loading`/`disabled` state:

```sass
border-top-color: $color-black--400;
border-right-color: $color-black--400;
border-bottom-color: $color-black--400;
```

illusivesunrae:
There has been no movement on this PR in 3 months. Closing, but feel free to submit PRs in the future with the fixes described above.

PR #69, opened 2019-02-14
Add specific bug and feature issue templates

levimcg:
This PR adds issue templates for specific issue types. I've added templates for bugs, and feature requests.

I wonder if a third _discussion_ or _question_ template might be useful? We could use the _about_ field in the issue template to try and direct folks looking to ask question to the a Slack channel, etc. I've not put a lot of though into it, but seems like it could be useful.


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request
+about: Suggest an idea for this project
+---
+## Summary
+Add a clear and concise description of the problem you are trying to solve with this new feature. For example:
+
+> I have the following use case, but there's nothing in Rivet to support it [...]
+
+## Describe the solution you'd like
+Add a clear and specific description of what you'd like to see enhanced in or added to the Rivet.
```
illusivesunrae:
```suggestion
Add a clear and specific description of what you'd like to see enhanced in or added to Rivet.
```

levimcg:
@illusivesunrae Those are good questions. Yes, I would prefer that we direct folks to the Slack, email, etc. for questions so as not muddy up the backlog/issue tracker.

We have created a `discussion` tag for the issues that do get created that are somewhere between requests and questions. Maybe that label is good enough for the issue that do get create that don't fall into the _bug_ or _feature request_ category.

Issue #51, opened 2019-01-17
Update to Gulp 4.0


levimcg:
I'd like to get Gulp updated to 4.0. 4.0 was a pretty major/breaking update to Gulp so it will require rewriting and/or refactoring a lot of the tasks we have written. Most notable is that you can now choose to run tasks in [series](https://gulpjs.com/docs/en/api/series) or [parallel](https://gulpjs.com/docs/en/api/parallel) which is a pretty big change.

https://gulpjs.com/docs/en/getting-started/quick-start

nbannist:
I think > 95% of it is converting to using the series/parallel methods from the array of tasks used in previous versions. I'm using in for the new kb styles.

PR #70, opened 2019-02-14
Feature/dependencies

illusivesunrae:
Addresses issue #51, and updates critical and severe dependencies.


config/css.js
```diff
@@ -41,6 +40,6 @@ gulp.task('css:prefix-release', function () {
     .pipe(gulp.dest('./css/'));
 });
 
-gulp.task('css:release', function (done) {
```
levimcg: I noticed the `css/` folder with the version exapanded and minified version of `rivet.css` weren't getting built into the root of the directory when running `npm run build`. Which also means that there's no `css` folder added to the final `rivet.zip` that gets built.

You can get rid of this callback function here (`function(done)`) and it will fix it. So it would become:

```javascript
gulp.task('css:release', gulp.series(
  'css:dist',
  'css:prefix-release',
  'css:header',
  'css:minify'
));
```
Just like you have the javascript `js:release` task set up below.

config/javascript.js
```diff
@@ -42,22 +41,20 @@ gulp.task('js:vendor', function() {
  * Watch scripts for changes and move to the "static" folder
  */
 
-gulp.task('js:watch', function() {
-  gulp.watch('src/js/**/*.js', ['js:concat', 'js:vendor']);
+gulp.task('js:watch', async function() {
```
levimcg: I'm pretty sure we're good on the use of `async` anywhere we build these files. I looks like default support for `async`/`await` was added in node 8.x. Just something to think about as we're setting up new CI stuff. We'll need a minimum of of 8.x version to build.

As an alternative you can also pass a function to the call back here and eliminate the need for using async:

```js
gulp.task('js:watch', function(callback) {
  // all you task stuff here...

  callback();
}

File: config/javascript.js [hunk 3]
```diff
@@ -42,22 +41,20 @@ gulp.task('js:vendor', function() {
...
 });
 
 gulp.task('js:dist', function() {
   return gulp.src('static/js/rivet.js').pipe(gulp.dest('./js'));
 });
 
-gulp.task('js:header', function() {
-  gulp
-    .src('./js/rivet.js')
+gulp.task('js:header', async function() {
```
levimcg: Same comment as above here. I could see the benefit of just using regular old callbacks consistently instead of sprinkles of async code. Just an idea though, maybe a little nitpick-y 😄

gulpfile.js
```diff
@@ -12,42 +11,39 @@ requireDir('./config');
  */
 
 gulp.task('build', function(cb) {
-  runSequence('sass', 'js:concat', 'js:vendor', 'fractal:build', 'css:prefix-fractal', cb);
+  gulp.series('sass', 'js:concat', 'js:vendor', 'fractal:build', 'css:prefix-fractal', cb);
```
levimcg: Not sure you need this `cb` argument here. that was used as a callback from the `runSequence` plugin that we're no longer using. I would definitely test that though.

levimcg:
@illusivesunrae Oh, also, just a heads up. You'll need to pull in `develop` before we merge.

PR #75, opened 2019-02-21
Added new flexbox utility classes

levimcg:
This PR adds a new set of comprehensive flex utilities to help developers have more fine-grained control over layout. White the grid is good for overall page layout, things can get messy when trying to nest too many grids to achieve complex layouts.

This new set of flex utilities gives developers a set of CSS classes they can combine to create more complex responsive layouts at the component level.

This adds utility classes for all of the major flexbox properties that affect **flex containers**:

- `display`
- `flex-direction`
- `flex-wrap`
- `justify-content`
- `align-items`
- `align-content`

And some that apply to **flex items**:

- `flex-shrink`
- `flex-grow`
- `align-self`

Flex properties **not included** in these utilities (container and items):

- `flex-flow` - This container property is just shorthand for `flex-direction` and `flex-wrap` no need to repeat
- `flex-basis` - This property is used to set default widths of flex items. We already have the grid for defined widths.
- `order` - this is a hard one to standardize as we don't know how many flex items will be in any given flex container.

I've added some basic examples, but I'll plan to write more comprehensive docs when I update the docs site before the February release.

### DIFF
--- a/src/components/100-utilities/config.yml
+++ b/src/components/100-utilities/config.yml
@@ -1,1 +1,0 @@
-status: "ready"
--- a/src/components/100-utilities/utilities--flex.hbs
+++ b/src/components/100-utilities/utilities--flex.hbs
@@ -1,0 +1,56 @@
+<div class="rvt-flex">
+  {{# each demos }}
+    <div class="flex-demo">
+      <div class="flex-demo-inner">Item {{ this }}</div>
+    </div>
+  {{/each}}
+</div>
+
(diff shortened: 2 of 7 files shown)


scottanthonymurray:
I'm excited about these! Let me know how I can help with the documentation, I know the biggest thing will be helping developers differentiate between when they should use the grid for layout and when they should reach for these utility classes.

levimcg:
@scottanthonymurray Yeah, for sure. I'll take a crack at putting a first version of the docs together and then tag you for a review/help. 🙌

PR #107, opened 2019-04-15
Added responsive width utility classes

scottanthonymurray:


src/sass/utilities/_utilities-width.scss
```diff
@@ -0,0 +1,36 @@
...
+  lg: $width-base * 5,         // 320px
+  xl: $width-base * 6,         // 384px
+  xxl: $width-base * 8         // 512px
+) !default;
+
+// Builds non-responsive width utility classes
+
+@each $size, $width in $width-map {
+  .#{$prefix}-width-#{$size} {
+    width: $width;
```
levimcg: I think what might be better here is to set the `width` property to `100%` and then keep `max-width: $width`. That way the width of the element will be constrained to the value of `$width`, but will also be responsive (shrink to fit) if the parent element of whatever this width utility is being used on becomes smaller than the `max-width`.

levimcg: I think this ☝️ will be the last thing before this is done. So this would become:

```sass
.#{$prefix}-width-#{$size} {
  width: 100%;
  max-width: $width;
}

PR #113, opened 2019-05-17
Added new menu component markup, CSS, and docs outline

levimcg:
This PR addresses #3 

We've been talking about this one for a long time now, and it's turned out to be pretty complex. 🤯

This PR specifically addresses the need for a flexible navigation or menu component that can be used outside of the header/global navigation of an application or site

Long-term, it seems like it would make sense to combine this _menu_ visual style with the current [Rivet tabs component styles](https://rivet.iu.edu/components/page-content/tabs/). We should very clearly document that although the visual styles are the same, the implementation is very different.

For example, we need to be **very** clear about the difference between changing routes/urls and swapping sections of content on the same page.

I've added bullets to the README here to outline some initial docs, but they need a bit of work before the next release.

CC: @christiemh54 @basham


scottanthonymurray:
@levimcg Another exciting addition to Rivet!

I took a look in Fractal and didn't see any issues with the visuals.

My initial feeling when looking at the implementation was that it was strange for the wide version to be the special case. In my mind, the wide version that collapses into a vertical list makes more sense as the default, with a modifier class forcing a vertical display. This might be because I've just come off the step indicator where this is the approach.

I also feel like the modifier class `--responsive` is unclear and can imply that the other implementation is not suitable for mobile. I might suggest having the wide version that collapses be the default, and then a `.rvt-menu--vertical` modifier to force the vertical display at all screen sizes, not unlike the process step indicator.

I'm happy to open a PR to this effect. :)

levimcg:
I like this idea a lot. I felt a bit uneasy about the `--responsive` modifier as well—naming things hard. The goal was basically what you describe here, a version that stays vertically stacked, or can display inline. I can see folks wanting/needing both, but didn't want to force the responsive behavior by automatically building it in because there are some instances where the user of this component might want it to display inline on small screens as well.

I think the solution you're proposing here is good. No need to open a separate PR, if you want to implement the naming conventions and functionality on this branch that would be great. 👍

scottanthonymurray:
@levimcg Alright, I've modified the component in the following ways:

* The default `.rvt-menu` stacks vertically on small screens and lays out horizontally on larger screens
* You can apply the `.rvt-menu--vertical` modifier class to force vertical stacking on all screen sizes
* You can apply the `.rvt-menu--horizontal` modifier to force horizontal stacking on all screen sizes

levimcg:
Nice, I like this solution. Seems to make the functionality _much_ clearer. Nice work. 👍

src/sass/components/_menu.scss
```diff
@@ -0,0 +1,86 @@
...
+    #{$selector} &__item a:hover::after,
+    #{$selector} &__item button:hover::after {
+      background-color: $color-black--200;
+    }
+  }
+}
+
+@include horizontal-menu('&--horizontal');
+
+@include mq(bp(md)) {
```
levimcg: One thing here. A while back we added regular variables for all the breakpoints. I'd kinda like to move away from using this `bp()` function that gets values out of the main `$breakpoints` map and instead use the regular named variables. So in this case `$breakpoint-md`. Makes things a little more understandable, I think.

scottanthonymurray: Sounds good to me! Would you like for me to go through and replace all instances of the `bp()` function with the breakpoint variable reference as a separate branch for the next release?

levimcg: Yeah, if you have time that would be great. I know there are some straggler instances of it out there. 👍

scottanthonymurray: I'll file an issue and get on it, thanks!

Issue #35, opened 2019-01-17
Process steps completed indicator. 


levimcg:
Something that is a like the progress bar in ApplyIU, that can be used/configured similar to the breadcrumb React component. Ideally it could be styled for vertical or horizontal use.

krishancock:
I'm working on something like this for ESS. Here's a link to the design:

https://tt6lt5.axshare.com

wbobe:
The last two years we did two different methods of handling this for Statewide. In 2017 Patrick built something very similar to the ApplyIU one. Last year I think I just used Rivet breadcrumbs. I could show you examples of both of the solutions, if you want.

jameswilson:
This would be an interesting feature to standardize.  The Libraries website — currently using framework, not rivets — implemented the following solution:

https://libraries.indiana.edu/request-educational-access-library-subscription-electronic-resources-another-campus
https://libraries.indiana.edu/rush-cataloging-request

levimcg:
@wbobe @jameswilson Awesome, thanks for sharing. This was actually one of the projects we used in our first community workshop we had just a couple of weeks ago. We have a branch open right now that is based off of the work that the team in the workshop did:

https://github.com/indiana-university/rivet-source/tree/feature/step-indicator

We had hoped to get it into our latest release (yesterday), but weren't able to get testing and documentation done in time. But, it's looking like we're in pretty good shape to get it into the release next month!

illusivesunrae:
Pulled into master with PR #117.

PR #114, opened 2019-05-20
Added new step indicator component

levimcg:
This PR adds a new step indicator component based on some work that came out of the first Rivet community workshop.

See #35 for background information and discussion. Documentation is in progress on a new branch on [rivet-docs](https://github.com/indiana-university/rivet-docs/tree/feature/step-indicator).


scottanthonymurray:
@levimcg So happy to see this come together!

Code itself looks good, just a few questions about implementation I want to be sure about before editing the docs and examples in both this PR and `rivet-docs:step-indicator`:

* Should the current step be indicated with `aria-current="page"` or `aria-current="step"`?
* The first example in this PR suggests that all step indicators be hidden with `aria-hidden`, is this correct? It sounds right to me.
* If that's the case, is it necessary to put the `<span class=".rvt-sr-only">Step</span>` inside the step indicator, as the screen reader will never get there?

levimcg:
> Should the current step be indicated with aria-current="page" or aria-current="step"?

Good call on this one. I guess I'm' not 100% sure. I tested both in VoiceOver on Mac, and unfortunately, `aria-current` had no affect, which will have some affects on other places we are using it to indicate the current page (e.g. the header). Not saying that we shouldn't recommend using `aria-current` for the assistive technologies that do properly implement that attribute though. It's a weird situation here because it kinda indicates both the current `step` and/or `page` (new URL).

> The first example in this PR suggests that all step indicators be hidden with aria-hidden, is this correct? It sounds right to me.

Hmmm? I'm sorry, I may have written that down at some point without giving it too much though. It seems to me that `__indicator` element should be used to visually indicate the state of progress, so the number of the step, or the successful (or not) completion of the step, and the `__label` element should be used to communicate the name of the step as well as its state (current, errors, etc.) to assistive technology. I'm not 100% sure the best recommendation, but what we want to avoid is having a screen reader, for example, read out something like "**one**", "step **one**". Where the first "one" is the visual indicator (content inside the circle). So I would think it makes sense to hide the content in the `__indicator` element (`aria-hidden`) and use some combination of dynamically updating visually hidden content inside of the `__label` element (maybe with `.rvt-sr-only`), or the use of `aria-label`? I can take a look at this tomorrow and see if I can better write out these thoughts and do a bit of testing.

> If that's the case, is it necessary to put the <span class=".rvt-sr-only">Step</span> inside the step indicator, as the screen reader will never get there?

See above ☝️ on this one. Let me think on this a bit more and maybe test it out a bit. So much of this is dependent on the type of content, etc. So it my end up that we need to make a handful of recommendations on how properly label things.

scottanthonymurray:
@levimcg Sure thing, thanks for your insights! I'll sit on the docs for a day and circle back tomorrow. I can start on the menu component docs in the meantime.

levimcg:
@scottanthonymurray Just had a chance to briefly check out the branch you have open on rivet-docs for this and it looks great. I think next week before the release I can help finish up the a11y docs/recommendation in terms of labeling, etc.

I think this branch is in a good place, so I'm going to finish it up. 👍

Issue #111, opened 2019-05-14
Focus styling issue on input groups


thrclark:
There's a focus styling issue on input groups that becomes very apparent when the secondary button style is applied:

![Screen Shot 2019-05-14 at 3 02 08 PM](https://user-images.githubusercontent.com/5652113/57726858-0ededa00-765e-11e9-9fd2-7f52f0ed11b9.png)

We were able to hack around it with this:

`.rvt-input-group__input:focus:not(:last-child) {
	z-index: 201;
}`

But there's probably a better way to integrate a solution within the .scss structure.

levimcg:
@thrclark Thanks for opening this. This seems like a good update to make. We've pulled it into the backlog. 👍

illusivesunrae:
Released in PR #140.

PR #123, opened 2019-07-10
Updated focus for input groups

illusivesunrae:
Fixed bug related to z-indexes of elements within input groups. For more information see issue #111.

### DIFF
--- a/src/sass/components/_input-group.scss
+++ b/src/sass/components/_input-group.scss
@@ -18,8 +18,9 @@
     z-index: map-get($z-index, z-100);
   }
 
-  &__input:focus:not(:last-child) {
-    z-index: 201;
+  &__input:focus:not(:last-child),
+  &__input:focus:not(:first-child) {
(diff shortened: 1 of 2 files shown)


src/sass/components/_input-group.scss
```diff
@@ -18,6 +18,10 @@
     z-index: map-get($z-index, z-100);
   }
 
+  &__input:focus:not(:last-child) {
+    z-index: 201;
```
levimcg: Cool, I think this looks good. One quick thing. We do have a Sass map for `z-index` values that might be good to use here instead of a hard-coded value, assuming there isn't a good reason for hard-coding. So instead of `201` maybe:

```scss
&__input:focus:not(:last-child) {
  z-index: map-get($z-index, z-300);
}
```

illusivesunrae: Oh nice! I definitely should've noticed that.

Issue #18, opened 2019-01-17
Feature request: Timeline


scottanthonymurray:
A component used to show historical interaction within a system, currently being used in IU Classifieds Administration to view an individual user's interaction of ad postings:

http://www.indiana.edu/~espd/rvb/develop/php/lib-extra-timeline.php

illusivesunrae:
I went back and investigated the timeline that we incorporated into the OVPUE version of the framework during my time there, and found that we used the [CodyHouse Timeline](https://codyhouse.co/gem/vertical-timeline/) as the base. An example of the OVPUE framework timeline is available on the [Wells Scholars Program website](https://wellsscholars.indiana.edu/deadlines/office-admissions.html).

illusivesunrae:
Released in PR #140.

PR #135, opened 2019-07-16
Added new Timeline component

illusivesunrae:
This PR adds a new Timeline component. See issue #18 for more information.


levimcg:
@illusivesunrae This is looking really good, very nice work! On the back of our face-to-face conversations this morning, just wanted to get my notes here.

I'm happy to work through this stuff 1:1 and review once you've had a chance to do some refactoring. If you're able to pull in @zttodd and maybe to dome remote pairing as you're refactoring, that might be nice. Any opportunity for him to get some more experience in the code base would be good right now.

Once you've had a chance to take another pass at it based on this feedback, we can circle back. 👍 

## Initial thoughts
- Hard-coded values should be minimized and should always come with a comment explaining why
- Components shouldn't have any outer margin
- BEM modifiers `.rvt-timeline--right` instead of `.rvt-timeline-right`
- Remove unused data attributes in markup examples
- Use the `mq()` mixin as much as possible see `_tools.scss` for more info
- Add `aria-hidden` to `__marker` as it's purely presentational and has not content inside.
- We shouldn't use utils on component elements when they already have an _element_ (e.g. `rvt-timeline__content`) class name applied
- Dates/labels on right and center-aligned variants rely on hard-coded magic numbers and fall apart if the outer container is not `--junior` These should work without any knowledge of the parent container's max width. Will need to use some combination of absolute positioning
- Generally speaking we should think about the `__content` area as a blank slate. We should show examples with good defaults (like what you have with util classes), but should design for the reality that lots of different stuff could possibly end up in that container. That said the example content you've put inside the `__content` area does a nice job of demonstrating how you could use type and spacing utils to build a nice layout.

Thanks again for taking this on. This'll be a great addition once it's all finished up. 💯

levimcg:
Just an FYI, I've made some notes on this and have time blocked on my calendar first thing tomorrow morning to do another review.

src/components/29-timeline/timeline--centered-dates.hbs
```diff
@@ -0,0 +1,24 @@
+<div class="rvt-timeline rvt-timeline--center">
+  <div>
+    <div class="rvt-timeline__item">
+      <div class="rvt-timeline__marker" aria-hidden>
```
levimcg: `aria-hidden` needs to have a value set of `true`. Even thought the absence of the the `aria-hidden` attribute is the same as `araia-hidden="false`, `aria-hidden` with no value is not the equivalent of `aria-hidden="true"` (like say `disabled` or `hidden`) as far as browser a11y APIs are concerned. This article has a pretty detailed explanation of the differences.

https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

This goes for all instances of `aria-hidden` in the PR

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+  &__content {
+    padding: $md;
+    margin-left: $lg;
+    background-color: $color-white;
+    border-radius: $xxs;
+    box-shadow: $shadow-base;
+    border: none;
+  }
+
+  & h2 {
```
levimcg: Generally speaking, if you see generic element selectors inside a component that means the style is too tightly coupled to some specific markup. In this case, the component will only work if the heading happens to be an `h2` element. Depending on the site/application an `h2` might not be correct in the page outline.

Since we're already adding a `rvt-timeline__date` element inside here you could do something like `rvt-timeline__heading` for what is currently an `h2`. That way the developer can use what ever heading level they need to use and the styling can remain consistent.

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+      border-right-color: $color-black--100;
+      border-left-color: transparent;
+    }
+
+    .#{$prefix}-timeline__marker {
+      margin-left: -$xxs;
+      left: 100%;
+    }
+
+    .#{$prefix}-timeline__item-right .#{$prefix}-timeline__marker {
```
levimcg: `#{$prefix}-timeline__item-right` and `#{$prefix}-timeline__item-left` should be modifiers if there is already an existing `#{$prefix}-timeline__item` element class that applies styles common to both the `--left` and `--right` variants.

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+      position: absolute;
+      display: block;
+      left: initial;
+      right: -$sm*7;
+      top: unset;
+      margin-top: -$sm*3.15;
+    }
+
+    .#{$prefix}-timeline__item-right .#{$prefix}-timeline__date--label {
+      min-width: fit-content;
```
levimcg: I'm pretty sure `fit-content` isn't supported in IE11. You could use `auto` here or some explicit value, but you'll need to test.

https://caniuse.com/#search=fit-content

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+      min-height: fit-content;
+      left: -$sm*7.5;
+      right: initial;
+      text-align: right;
+      position: absolute;
+      display: block;
+      top: $sm*.75;
+      margin-left: unset;
```
levimcg: `unset` isn't supported in IE11. You could use `initial` here or an explicit value like `0`. All instances of `initial` will need updated.

https://caniuse.com/#search=unset

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+      margin-top: unset;
+      margin-right: unset;
+    }
+  }
+}
+
+@include mq($breakpoint-xl) {
```
levimcg: We should make these labels display to the side at a smaller breakpoint. Maybe `$lg` or `$md`? I get the feeling that folks would expect the outset labels to display this way on more average sized screens as well. Right now the screen has to be quite large before you they display out to the side.

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+  .#{$prefix}-timeline {
+    &__date--label {
+      min-width: fit-content;
+      max-width: fit-content;
+      min-height: fit-content;
+      left: -$sm*7.5;
```
levimcg: Let's bump the distance between the center line and this date down a bit. It feels like there might be a little too much space between the timeline item and this date. Maybe something like `$sm*7`.

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+}
+
+@include mq($breakpoint-md) {
+  .#{$prefix}-timeline--center {
+    &::before {
+      left: 50%;
+      margin-left: 0;
+    }
+
+    & > div {
```
levimcg: This feels a bit off. Is there a specific reason not to set `--timeline-center` to `display: flex` and  then also set `__item` to `display: flex`? It is possible for an element to be both a flex container and a flex child. This would need some testing to confirm, but similar to my previous comment on super generic selectors, could we tie this styling to an explicit class name that is already being used in the component?

src/components/29-timeline/timeline--right-dates.hbs
```diff
@@ -0,0 +1,20 @@
+<div class="rvt-timeline rvt-timeline--right">
+  <div class="rvt-timeline__item">
+    <div class="rvt-timeline__marker" aria-hidden>
+    </div>
+    <div class="rvt-timeline__content">
+      <h2>Sample Item</h2>
+      <span class="rvt-timeline__date rvt-timeline__date--label">Spring 2019</span>
```
levimcg: On the second item in the timeline the date sets up against the item marker (blue dot). See attached screenshot.

image

src/sass/components/_timeline.scss
```diff
@@ -0,0 +1,215 @@
...
+      margin-top: unset;
+      margin-right: unset;
+    }
+  }
+}
+
+@include mq($breakpoint-xl) {
+  .#{$prefix}-timeline {
+    &__date--label {
+      min-width: fit-content;
```
levimcg: In order to accept longer pieces of text in the `__date` element. We'll need to set some sort of explicit width on larger screens and adjust the absolute positioning. That way if the text gets longer it will flow onto two lines. See screenshot of what currently happens:

image

We should be able to fix this by changing the `min-width` property here to be something like `10rem` (that's just an arbitrary number). Once we decide on a sensible default width, we'll need to adjust the `left` position to accommodate the new fixed width of the `&__date--label` element. We can take a look at this when we meet Friday morning if you'd like. 👍

src/sass/components/_timeline.scss
illusivesunrae: The reason I did it this way was to ensure that each item is in its own row, rather than stacking up together:

Screen Shot 2019-07-24 at 3 44 22 PM

PR #138, opened 2019-07-24
Refactor button component configuration and template

zttodd:
- Reorganizes `buttons.config.yml` to include type, secondary, and small variants to be used within the Handlebars templates.
- Refactors `buttons.hbs` to output each of the different variants as separate pages.


src/components/09-buttons/buttons.config.yml
```diff
@@ -1,8 +1,47 @@
...
+
+    - name: "secondary default"
+      context:
+          label: "Primary"
+          secondary: "secondary"
+    - name: "secondary success"
+      context:
+          type: "success"
+          label: "Success"
+          secondary: "secondary"
```
levimcg: Since a button is either secondary or not, what do you think about making it a boolean instead of a string? I see that you're using the value ("secondary") inside the template to apply the `--secondary` modifier class, but it might make more sense to check the truthiness of the `secondary` key and then apply the `--secondary` modifier based on that instead of interpolating the `{{ secondary }}` value in the template.

```yaml
secondary: true
```

src/components/09-buttons/buttons.hbs
```diff
@@ -1,87 +1,11 @@
...
-        </tr>
-    </tbody>
-</table>
+<button class="rvt-button
+  {{~#if type}}
+    {{~#if secondary}} rvt-button--{{type}}-{{secondary}}{{~/if}}
+  {{~/if}}
+  
+  {{~#if type}} rvt-button--{{type}}{{~/if}}
+  {{~#if secondary}} rvt-button--{{secondary}}{{~/if}}
```
levimcg: See my comment above. Just to clean up the config file and make it a bit more readable. Setting `secondary: true` in the config and then in the template here:

```handlebars
{{~# if secondary}} rvt-button--secondary{{~/if}}
```

src/components/09-buttons/buttons.config.yml
zttodd: I briefly considered this as well and wasn't sure if the 'secondary' value could potentially change in the future. Using a boolean makes sense! I'll make that update.

src/components/09-buttons/buttons.config.yml
```diff
@@ -1,8 +1,47 @@
+title: Buttons
 status: "ready"
+context:
+    label: "Primary"
```
illusivesunrae: I think in the context of a button we should consider updating `label` to `content`.

levimcg:
Oh cool. That's neat. That makes a lot of sense to me. I like the idea of building all the logic into the component template file like that so that we can just pass data into the references (like `props` in other component models) inside of other templates/components in Fractal.

zttodd:
The primary and secondary buttons are now switched over from Handlebars to Nunjucks, but might be able to be refactored into one template using the features of Nunjucks. I also worked through a first draft of refactoring the segmented buttons to include both primary, secondary, and fitted within one template.

PR #139, opened 2019-07-25
Refactored text inputs and textareas in fractal

illusivesunrae:
In this PR:

- Renamed components to singular `text input` and `textarea`
- Created new folder for `textarea`
- Updated functionality to pull appropriate inline alerts programmatically


illusivesunrae:
@levimcg I'm wondering if instead of having `text input` and `textarea` as separate components entirely, if it would be better to add them as separate subfolders within a larger folder `text fields` or something. Let me know what you think.

levimcg:
Hmmm? I think my preference would be to just have them as totally separate components unless you're thinking of some specific reason to not do that?

In my mind, from a dev/user's perspective, it helps to remove any ambiguity. For example, if you know that you are specifically looking for a `textarea` element you should be able to see that in a list of components instead of guessing whether or not they are grouped with something like _Text inputs_.

I'd like to move this direction in the documentation, so I think it make sense to do that here as well. Does that make sense?

src/components/05-text-input/text-input.hbs
```diff
@@ -0,0 +1,5 @@
+<label for="demo-1">Text Input</label>
```
levimcg: Let's make this a prop you can pass to the `context` object instead of hard coding an `id`. That way we can use more than one per page with unique label/input pairs.

src/components/30-textarea/textarea.hbs
```diff
@@ -0,0 +1,5 @@
+<label for="demo-{{name}}">Textarea</label>
```
levimcg: I'd like to see this same template logic applied to the `text-input.hbs` template (see previous comment).

src/components/30-textarea/textarea.config.yml
```diff
@@ -0,0 +1,30 @@
+title: "Textarea"
+status: "ready"
+collated: true
+context:
+    alert: false
+variants:
+    - name: "disabled"
```
levimcg: Once you've made the updates to the templates, we'll need to add `id` fields to all of the `context` objects.

src/components/05-text-input/text-input.config.yml
```diff
@@ -0,0 +1,30 @@
+title: "Text input"
+status: "ready"
+collated: true
+context:
```
levimcg: Same goes for this config file. Once the template is updated to accept and `id` prop, we'll need to add them to all of the `context` objects in the config.

PR #144, opened 2019-08-05
Refactored select inputs in fractal

illusivesunrae:
In this PR:

- Updated functionality to pull appropriate inline alerts programmatically
- Refactored code to allow replacement options within other examples using an array (`options`)


zttodd:
I got that warning as well, but without saving any changes.

Otherwise, it looks good!

illusivesunrae:
I've actually had this pop up on all my branches since we started the fractal restructure. I think it should be fine for now, but we should re-assess, maybe check with the Fractal slack?

PR #146, opened 2019-08-07
Switch templating engine from Handlebars to Nunjucks

zttodd:
- Switches the templating engine in `fractal.js` from the default Handlebars to Nunjucks (`.njk`).
- Updates `_preview.njk` to Nunjucks syntax.
- Updates the alert component and its variants to provide an example of the new syntax.
- Updates the syntax for alert includes within `101-includes` so the default alert and its variants will work.


zttodd:
@illusivesunrae I was able to fix the nested include issue by changing from the Handlebars' helper `view` to the Nunjucks' `include` tag for the header include within an include. The header is outputting the different variations correctly now.

illusivesunrae:
Awesome! 💯

PR #150, opened 2019-08-09
Refactored list component and converted to Nunjucks

zttodd:
Originally while refactoring the list component, I discussed with @illusivesunrae whether the individual list variants should each be within its own directory. We decided they would, because the template code for including all of the variants within one file would not be human friendly to read and work with.

Separate from this refactor, I was thinking of creating a new `data-include` component which will consist of data lists, objects, arrays, etc to include within different components, because there is a lot of duplication of the same placeholder data happening between components. If we go that route, this component and the button component refactor will be updated to pull from that data source instead.


levimcg:
Hmm, that's a cool idea, having some shared data to use sort of _globally_. I _think_ you can add data like that to the root `components.config.yml` file. Then you can access from inside any component.

I put some data in there at some point and was using in the component templates:

https://github.com/indiana-university/rivet-source/blob/develop/src/components/components.config.yml

Might be something worth digging into.

zttodd:
Ah yes, that's exactly what I was thinking. I'll look into including the placeholder data there instead. Thanks!

PR #151, opened 2019-08-12
Moved global data into components.config.yml

zttodd:
This moves global data that can be used across components and variants into the `components.config.yml` file. An example that's used within the `list` component:

```
list_items:
    - List Item One
    - List Item Two
    - List Item Three
    - List Item Four
```

Other related fixes:
- Adjusted indentation inside components.config.yml.
- Removed 'collated: true' from list component variants since it's not needed in this case.

### DIFF
--- a/src/components/04-list/list-plain-text/list-plain-text.njk
+++ b/src/components/04-list/list-plain-text/list-plain-text.njk
@@ -1,5 +1,5 @@
 <ul class="rvt-plain-list">
-  {% for text in texts %}
-    <li>{{ text }}</li>
+  {% for list_item in list_items %}
+    <li>{{ list_item }}</li>
   {% endfor %}
 </ul>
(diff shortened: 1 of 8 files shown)


src/components/04-list/list-plain-text/list-plain-text.njk
```diff
@@ -1,5 +1,5 @@
 <ul class="rvt-plain-list">
```
illusivesunrae: I'm wondering if it might be better to rewrite the `list-plain-text` to be a variant of `list-unordered` since the only difference is the presence of the class `.rvt-plain-list`.

zttodd: That's a good idea. I can include the `inline` list variant within that as well, since that was left out in the list component refactor. That can be in a separate branch.

PR #154, opened 2019-08-12
Refactored select-input component to Nunjucks

zttodd:
Since you were working on `text-input` I started on this one. It should be working now, and I think we _might_ be able to use this solution with the `text-input` component as well. The main issue in this case ended up being that the `alertVariant` required `includes--` at the beginning in the config file, and each of the values' filenames had an incorrect spelling. I was doing all sorts of trial and error with the `include` and the values and context in the config file and eventually realized that that was the issue. I took some notes through the debugging that we can go over on Wednesday, and we might want to double-check the 101-include filenames to make sure they didn't get renamed at some point during this refactor as well.


zttodd:
This select-input code from the initial commit to this branch has been updated to include `alert-inline` variants, instead of pointing directly to the icons within `101-includes` and duplicating wrapper code.

PR #156, opened 2019-08-14
Refactored text-input

illusivesunrae:
Updated the file extensions to `.njk` and refactored the text-input and alert-inline components.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -959,50 +959,103 @@
       }
     },
     "@frctl/fractal": {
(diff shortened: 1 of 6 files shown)


zttodd:
My only comment is to add spaces around the two variables "id" in `src/components/05-text-input/text-input.njk`, so that it's `{{ id }}`. Other than that, the logic for this looks really clean and I'm going to use it for the text-input component as well.

PR #177, opened 2019-08-22
Refactored modal component

zttodd:
Notes for this PR:

- The close button icon was copied into a separate include, which was then included within the `alert-close-button` include.
- For the attributes on the outer `<div class="rvt-modal">`, a combination of configuration values and template hard-coded values were used, instead of using an `attributes` list within the `config.yml`. This is because there are several attributes that won't change across variants (aria-hidden, role, tabindex), so it will be less to have to copy/paste within the configuration and when rendering the component.
- There was a class of `v-hide` on the modal title in the confirmation dialog example, but that has been removed for this PR since it was breaking the layout for the close button.
- The custom JS was left in the template to be re-evaluated at a later date.


src/components/12-modal/modal.njk
```diff
@@ -0,0 +1,93 @@
+<button class="rvt-button" data-modal-trigger="{{ id }}">{{ buttonText }}</button>
```
illusivesunrae: `{% render '@button', { content: buttonText, attributes: [{ label: "data-modal-trigger", value: id }]}, true %}`

src/components/12-modal/modal.njk
```diff
@@ -0,0 +1,93 @@
...
+
+<div class="rvt-modal"
+  id="{{ id }}"
+  role="dialog"
+  aria-labelledby="{{ modalTitleId }}"
+  aria-hidden="true"
+  {% if ariaDescribedby %}aria-describedby="{{ ariaDescribedby }}"{% endif %}
```
illusivesunrae: For ones such as this, you might consider using `{% for attribute in attributes %}` logic instead and just adding them to the yaml file as
```
attributes:
    - label: "ariaDescribedby"
       value: "whateverValueYouWant"
    - label: "data-modal-dialog"
       value: "true"
```

src/components/12-modal/modal.njk
zttodd: Fixed!

src/components/12-modal/modal.njk
zttodd: The `aria-describedby` and `data-modal-dialog` attributes are now moved into a for loop.

Issue #13, opened 2019-01-17
Reverse-tabbing on new modal window


scottanthonymurray:
When a modal is opened, focus should shift to it and tab path should be contained within the elements of that modal; and this is the way that Rivel modal behaves when you forward-tab after opening. Howvever, if you reverse-tab after opening the modal. it will cause the focus to jump out of the modal and on to the page below. See the attached screenshot where the focus has been moved to the main screen (highlighted in yellow).

![screenshot_2](https://media.github.iu.edu/user/4398/files/0ae32f88-cbea-11e8-9f1e-e07247e28b17)

illusivesunrae:
Merged into master in PR #207.

PR #195, opened 2019-09-18
Fixed backward tabbing issue for modals

zttodd:
When initially opening a modal, if the user presses shift + tab, the focus leaves the modal. This fix sets the focus to the first focusable element within the modal when shift + tab is pressed upon first opening the modal.

For more details, see issue #13.

### DIFF
--- a/src/js/components/modal.js
+++ b/src/js/components/modal.js
@@ -260,7 +260,7 @@ var Modal = (function() {
       last.focus();
     } else if (document.activeElement === activeModal) {
       event.preventDefault();
-      first.focus();
+      last.focus();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/js/components/modal.js
```diff
@@ -258,6 +258,9 @@ var Modal = (function() {
     if (document.activeElement === first) {
       event.preventDefault();
       last.focus();
+    } else if (document.activeElement === activeModal) {
+      event.preventDefault();
+      first.focus();
```
levimcg: I would say it makes more sense to move focus to the last focusable element when the user presses `Shift + Tab` first.  So `last.focus();` instead.

zttodd: This has been updated 👍

PR #199, opened 2019-09-25
Fixed header drawer bottom border issue

illusivesunrae:
As described in issue #9, there is an issue with the display of a single identity module task in the mobile dropdown. This issue is actually two-fold.

In the example provided in the issue, the developer was likely using the incorrect mark up (though with reason). On the [Rivet Docs website](https://rivet.iu.edu/components/navigation/header/), we say "when there is only one user function consider replacing the carat with a bar followed by the single function", however we do not show any examples of a single user function alongside navigation items. When navigation is present, it makes sense (to a degree) to save space by using the carat even when only a single user function is present, but if this is preferred, it is not stated in the documentation.

The issue lies in the fact that the ["Header with identity menu" (single user function)](https://rivetdev.uits.iu.edu/develop/components/detail/header--id-module.html) and the ["Header with main navigation" (multiple user functions)](https://rivetdev.uits.iu.edu/develop/components/detail/header--navigation-identity.html) have different mark up for the identity modules in the drawer area. For the first example, the identity module portion is inside a div and is not encased in a `<nav>` tag; whereas in the second example, the identity module appears within the `<nav>` tag alongside the traditional navigation.

If one uses the combination of the single identity module mark up, but also adds navigation, the identity module in the mobile drawer does not receive a border-bottom, thus creating the problem shown in the issue example.

To resolve this I have created an example featuring a header with navigation and a single user function within the identity module area. The result within the drawer area was:
```
<li>
    <div class="rvt-header-id__profile rvt-header-id__profile--drawer">
        <span class="rvt-header-id__avatar" aria-hidden="true">rs</span>
        <span class="rvt-header-id__user">rswanson</span>
        <a href="#0" class="rvt-header-id__log-out">Log out</a>
    </div>
</li>
```

This mostly worked, but did experience specificity issues related to `.rvt-drawer .rvt-drawer__nav ul li a` styling overruling `.rvt-header-id rvt-header-id__log-out` styling. I have added a correction for this as well.

## Things that need to be decided:

- Should developers be encouraged to combine navigation with the single user function in the header, or should they be encouraged to use the carat-style dropdown (even when there is only a single user function) for the identity module when combined with navigation?
- If the dropdown is encouraged for a single item, should the mobile drawer use the single item markup shown above, or continue to use the mobile button toggle?


levimcg:
This is on my to-do list to review. I need to block off some time for it, but it may be tomorrow (09/27) before I can get to it.

Issue #212, opened 2019-11-07
Remove deprecated code


levimcg:
Now that we are going to begin development on the Rivet 2, we need to evaluate the current code base to look for any deprecated code that can be removed.

Most of the deprecated code will be isolated to the Sass files. I've done my best to mark anything that should be removed in the next breaking changes with a `DEPRECATED:` comment, so searching the code base to find things that need removed should be pretty straightforward.

Things to look for are old un-prefixed css classes. the first couple minor version of Rivet did not namespace the CSS classes with `rvt` so there are a lot of places where you'll see duplicated CSS classes in declarations. For example:

```scss
.text-bold,
.#{$prefix}-text-bold {
  font-weight: 700;
}
```

In these cases we need to remove the older un-prefixed versions where they exist.

@illusivesunrae @zttodd I'm happy to get together to chat about this in person if there are any questions.

zttodd:
After working through removing deprecated and non-prefixed selectors within `src/sass`, these files will need further review to evaluate how to to handle more general non-prefixed selectors.

`components`
- [ ] _checkboxes.scss
- [ ] _inputs.scss
- [ ] _radios.scss
- [ ] _tables.scss

`core`
- [ ] _base.scss
- [ ] _tools.scss

illusivesunrae:
Published to `2.0.0-alpha` with Release [v2.0.0-alpha](https://github.com/indiana-university/rivet-source/releases/tag/v2.0.0-alpha).

PR #215, opened 2019-11-08
Removing deprecated code

zttodd:
In this PR:

- Removed deprecated CSS and JS
- Removed duplicate, non-prefixed selectors throughout Sass files
- Removed `panel` component

See issue: #212 for more details.


src/sass/components/_links.scss
```diff
@@ -1,7 +1,6 @@
 // Copyright (C) 2018 The Trustees of Indiana University
 // SPDX-License-Identifier: BSD-3-Clause
 
-a,
```
levimcg: I would say that we might want to leave this for now until we figure how how we're going to approach scoping the css for general HTML elements. I believe this is the only place where the link color is defined right now, so removing the `a` selector from this list right now, might have some unintended consequences that are maybe outside the scope of this PR.

zttodd: @levimcg good call! I've put the `a` back into the `_links.scss` file.

PR #216, opened 2019-11-14
Updated utilities section

zttodd:
In this PR:

- Updated `utilities` files from `.hbs` to `.njk`
- Moved `type-scale` and `spacing` into the `utilities section
- Added `rvt-` prefix to utilities
- Removed deprecated `rvt-display-flex` and `rvt-vertical-center` code from `utilities--display.njk`


src/components/utilities/utilities.config.yml
```diff
@@ -1,8 +1,38 @@
...
-    - "Three"
-    - "Four"
-    - "Five"
+    flexText:
+        - "One"
+        - "Two"
+        - "Three"
+        - "Four"
+        - "Five"
+    spacings:
```
levimcg: What do you think about creating another object in here similar to this `spacings` object called something like `typeScale` where we can store the values of the type scale. That way we can do something similar in the `utilities--type-scale.njk` component file where we loop through and create the markup for the example. That way the values could be update in one single place if needed instead of digging through the hard-coded markup. Does that make sense?

illusivesunrae: Makes sense to me!

zttodd: Agreed! 👍

zttodd: Along with refactoring the Nunjucks and config.yml for `type-scale`, I've refactored each of the rest of the utilities as well.

PR #219, opened 2019-11-14
Added sidenav component

zttodd:
In this PR:

- Added the _Sidenav_ component
- Added `wip` (work in progress) status to `fractal.js` and set its color to orange
- Updated the color for the `deprecated` status to a red color value
- Updated the color for the `beta` status to the Rivet blue hex color value
- Updated the `aria-label` on `nav` from `Sidebar navigation` to `Sidenav`

JavaScript for this component will be added and refactored at a later time once the ES6 build process has been implemented.


fractal.js
```diff
@@ -65,7 +65,12 @@ fractal.components.set('statuses', {
     deprecated: {
         label: "Deprecated",
         description: "Don't use this in new projects. This component will be removed in the next major version of Rivet.",
-        color: '#F25B19'
+        color: '#FF0000'
+    },
+    wip: {
+        label: "Work in Progress",
```
levimcg: This is nice addition 😄

src/components/nested-navigation/nested-navigation.config.yml
```diff
@@ -0,0 +1,3 @@
+title: Nested Navigation
```
levimcg: After thinking about it I'm not sure if _Nested navigation_ is necessarily the most appropriate name. In theory, you could use this component as a side navigation that only had one level of non-nested links.

This is fine for now, but let's talk about his more on Monday.

src/sass/components/_nested-navigation.scss
```diff
@@ -0,0 +1,127 @@
...
+    border-top: 1px solid $color-black--100;
+    display: flex;
+    flex-wrap: wrap;
+    margin-top: 0; // resets Rivet default margin-top
+  }
+
+  &__link {
+    color: $color-black;
+    flex-grow: 1;
+    padding: $sm;
```
levimcg: The overall height of each nav item is looking a bit tall to me once you look at it in the browser. I bumped the `$sm` value down to a slightly small magic number and it felt much better:

```scss
padding: $sm*.875;
```

src/sass/components/_nested-navigation.scss
```diff
@@ -0,0 +1,127 @@
...
+      text-decoration: underline;
+    }
+  }
+
+  &__icon {
+    align-items: center;
+    background: transparent;
+    border: none;
+    justify-content: center;
+    padding: $sm;
```
levimcg: See comment above. You will need to adjust this value to match what ever reduced padding value we end up settling on for the link.

src/sass/components/_nested-navigation.scss
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    &:focus,
+    &:hover {
+      color: $color-blue;
+      text-decoration: underline;
+    }
+  }
+
+  &__icon {
```
levimcg: I wonder if it makes more sense to name this element `&__toggle`? `__icon` describes an element that is actually a child (the `svg`) of the element you are styling with this rule (a `button`).

src/sass/components/_nested-navigation.scss
```diff
@@ -0,0 +1,127 @@
...
+$list: 'rvt-nested-nav__list';
+$item: 'rvt-nested-nav__item';
+
+/* Set up initial selectors for indenting and shading */
+$indent-selector: '.#{$list} > .#{$item} > .#{$list} > .#{$item}';
+$shade-selector: '.#{$item} > .#{$list}';
+
+// Set up level indent/background shade map
+$level-styles: (
+  40: $color-black--050,
```
levimcg: Let's cut these indentation values down by half. When you think of how this component will generally be used as a sidenav/sidebar, in those deeper nested levels you might barely have enough width for one word, and if it's a long word you're going to run into one of those _CSS IS AWESOME_ situations where the text overflows outside of the link/item or else expands the width of the whole (`.rvt-nested-nav`) parent container.

src/sass/components/_nested-navigation.scss
zttodd: I agree! `&__toggle` makes more sense in this case.

src/components/nested-navigation/nested-navigation.config.yml
zttodd: We talked through this offline. We'll name this component  to _Sidebar_, since that will be its main intent.

illusivesunrae:
@levimcg, just took a look. This looks really great! I'll probably try to rework the `alert.js` on Monday to reflect this structure.

levimcg:
@illusivesunrae Perfect, sounds good!

PR #220, opened 2019-11-15
Added feature component

illusivesunrae:
In this PR:

- Brought over feature component markup and styling from codepen


src/sass/rivet.scss
```diff
@@ -24,6 +24,7 @@
 @import "components/buttons-segmented";
 @import "components/checkboxes";
 @import "components/dropdown";
+@import "components/features";
```
levimcg: Same as the comment above. Let's make all component references singular (`/feature`) to stay consistent.

illusivesunrae:
@zttodd, this is how it appears in the [figma file](https://www.figma.com/file/HTxdfdnZfyuVkppTuSDmAu/Rivet-2-Component-Specs?node-id=1%3A216).

src/sass/components/_feature.scss
```diff
@@ -2,40 +2,51 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 .#{$prefix}-feature {
-  max-width: 100%;
-
-  &__image {
-    overflow: hidden;
-  }
-
   & img {
```
levimcg: Since you're already wrapping the image in a div with the `rvt-feature__image` class. I think it makes more sense to move these declarations to a block using that class as a selctor. So instead of the more generic `img` selector it would become.

```scss
&__image img {
  display: block;
  width: 100%;
}
```

That way you're only targeting an `img` element within the wrapper instead of any image in the whole component, i.e. `.rvt-feature img`, which would target _any_ `img` that is a child of an element with `.rvt-feature` applied to it. Not that there _should_ be more images outside the `__image` wrapper, but just to be safe it makes sense to scope the styles you're writing with a little more intention in this case.

src/components/feature/feature.njk
```diff
@@ -6,15 +6,15 @@
...
     {% endif %}
-    <h2 class="rvt-feature__title">{{ title }}</h2>
+    <h2 class="rvt-feature__title">{% if titlelink %}<a href="{{ titlelink }}">{% endif %}{{ title }}{% if titlelink %}</a>{% endif %}</h2>
     {% if byline %}
       <p class="rvt-feature__byline">{{ byline }}</p>
     {% endif %}
     <p class="rvt-feature__teaser">{{ teaser }}</p>
     {% if meta %}
-      <p class="rvt-feature__meta">{{ meta }}</p>
+      <p class="rvt-feature__meta">{% if metalink %}<a href="{{ metalink }}">{% endif %}{{ meta }}{% if metalink %}</a>{% endif %}</p>
```
levimcg: I don't think this affects the CSS you've written at all, but I would imagine that any of the "meta" items here should be individual links that don't include the pipe as part of the link. So the markup should be able to be rendered as:

```html
<p class="rvt-feature__meta">
  <a href="#2">Meta text</a>  | <a href="#3">Sept. 16, 2019</a>
</p>
```

without the pipe being wrapped in a link.

It might mean making the nunjucks file accept an array of "link" objects and then looping through them to output separate link side the `rvt-feature__meta` element.

PR #235, opened 2019-12-02
Refactor color variables

zttodd:
In this PR:

- Refactored color hex values into a key/value sass map called `$colors`
- Used `map-get` to set `$colors` map key/values to variables

There are a bunch of comments in the `_variables.scss` file that can be cleaned up, but it looks like a lot will be taken care of with the black variant refactor.


src/sass/utilities/_utilities-color.scss
```diff
@@ -3,28 +3,17 @@
 
 /* stylelint-disable */
 
-@each $key, $value in $color-map {
```
levimcg: One more thing I forgot. You'll need to update the copyright year here to be 2019 since this is a new piece of code.

zttodd:
@levimcg I've added a `utilities--color.njk` example page for use with documentation and reviewing colors.

src/sass/utilities/_utilities-color.scss
zttodd: This file was around before these updates, previously using `$colors-map` to loop and produce background/foreground colors.

PR #237, opened 2019-12-03
Refactored alert js

illusivesunrae:
Refactored `alert.js` to reflect new js structure. Not currently passing any options--will need to discuss items that need to be added.

To Do:

- [x] Create `.remove` polyfill
- [x] Determine `options` to pass in


src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
 
-    this.init(options);
-  }
-
-  init(options) {
-    document.getElementById(options.id).closest('[aria-labelledby]').querySelector('[data-alert-close]').addEventListener('click', () => { this.handleClick(options.id); }, false);
-  }
+    // Instance properties
+    this.element = element;
+    this.closeAttribute = ['data-alert-close'];
```
levimcg: This is setting the `closeAttribute` class property to an array with one item (a string of `data-alert-close`) in it. 

It should be `this.closeAttribute = 'data-alert-close';` and then you can set `this.closeSelector = `[${this.closeAttribute}]` to the correct attribute CSS selector.

The reason you want the `this.closeAttribute` property to be only a string that is the data attribute name is that you may need to use it on it's own for some DOM methods e.g. `element.getAttribute(this.closeAttribute)`. In this case you con't need the fully-formed CSS selector string (with brackets `[]`), just the name of the attribute.

If you keep the attribute stored separately you can then interpolate it into the CSS selector you need later. e.g. `this.closeSelector = `[${this.closeAttribute}]` which gives you `[data-alert-close]` which you could use with `document.querySelector()`.

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
+    this.closeSelector = `[${this.closeAttribute}]`;
+    this.ariaAttribute = ['aria-labelledby'];
```
levimcg: You can get rid of these two properties. Instead of using `aria-labelledby` as a hook to grab the actual alert element, let's look at using a data-attribute to bring it inline with the rest of the components.

I'll add some more comments about this, stand by...

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
+    this.ariaSelector = `[${this.ariaAttribute}]`;
```
levimcg: We won't need this either.

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
 
     alert.remove();
+  }
 
-    if (callback && typeof callback === 'function') {
-      callback();
-    }
+  init() {
+    // Add click handlers
+    this.element.forEach((item) => {
```
levimcg: We'll need to rethink this strategy for attaching event handlers a bit. It looks like you are trying to look for and attaching  event listeners to multiple close buttons here, but there should only ever be one close button per alert.

What you'll need to do is query the main alert element (`this.element`) for the close button and then only add an event listener to that single close button inside the main alert element.

So this should become:

```js
init() {
  let dismissButton = this.element.querySelector(this.closeSelector);
  dismissButton.addEventListener('click', this._handleClick, false);
}
```

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
+      throw new Error(
+        'Couldn\'t locate the Aria selector for the element.'
+      );
     }
 
+    this.dismissAlert(dismissAria);
+
+  }
+
+  dismissAlert(dismissAria) {
```
levimcg: Based on my previous comments, this could then be simplified by doing a few things:

1. Remove the `dismissAria` parameter. You won't need to pass any arguments to this method
2. Get rid of the `let alert = dismissAria` statement
3. Change `alert.remove()` to `this.element.remove()`, we already have a reference in as a class property in the constructor to the actual alert element that we need to remove from the DOM.

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
+    let alert = dismissAria;
+
     if (!alert) {
       throw new Error(
-        'Could not find an alert with the id of ' + id + ' to dismiss.'
+        'Could not find an alert attached to the close selector to dismiss.'
       );
     }
 
     alert.remove();
```
levimcg: We may need to think about making a call to the `destroy()` method right before removing the alert just to clean up the event listener. I'm not 100% sure on this though, the event listener may get removed when the element is removed from the DOM via the `element.remove()` method. We'll need to confirm though.

src/js/components/alert.js
```diff
@@ -5,43 +5,67 @@
...
-    }
+  init() {
+    // Add click handlers
+    this.element.forEach((item) => {
+      let dismissButton = item.querySelector(this.closeSelector);
+      dismissButton.addEventListener('click', this._handleClick, false);
+    })
+  }
+
+  destroy() {
```
levimcg: You'll need to mimic the set up in the `init()` method when you refactor this based on the comments above.

cypress/integration/alert_spec.js
```diff
@@ -1,26 +1,39 @@
...
+    cy.get('@infoAlertClose').click();
+
+    cy.get('@infoAlert').should('not.exist');
   });
 
-  it('Should be able to close the alert', function() {
-    cy.get(ALERT_CLOSE).click();
+  it('Should be able to dismiss with .dismiss() method', function () {
+    cy.window().then(win => {
+
```
levimcg: You can get rid of these extra lines in here since this is the only statement inside this function body.

cypress/integration/alert_spec.js
```diff
@@ -1,26 +1,39 @@
...
+      win.newwarningAlert.dismiss();
+
```
levimcg: Same as above with this empty line

src/js/components/alert.js
```diff
@@ -2,46 +2,61 @@
...
-    const dismissButton = event.target.closest('[data-alert-close]');
+    // bind methods
+    this._handleClick = this._handleClick.bind(this);
 
-    // If the target wasn't the dismiss button bail.
-    if (!dismissButton) return;
+    // Check to make sure that a DOM element was passed in for initialization
+    if (!isNode(this.element)) {
+      throw new TypeError(
+        'A DOM element should be passed as the first argument to initialize the sidenav'
```
levimcg: This should be _...initialize the alert_ instead of _sidenav_.

src/js/components/alert.js
```diff
@@ -2,46 +2,61 @@
...
+  _handleClick() {
+    this.dismiss();
+  }
 
-    if (!alert) {
-      throw new Error(
-        'Could not find an alert with the id of ' + id + ' to dismiss.'
-      );
-    }
+  /**
```
levimcg: This comment should live inside the `dismiss()` method body next to the `dismissEvent` statement.

src/js/components/alert.js
```diff
@@ -2,46 +2,61 @@
...
+    if (!dismissEvent) return;
+
+    this.destroy();
+    this.element.remove();
+  }
 
-    alert.remove();
+  init() {
+    // Add click handlers
+    const dismissButton = this.element.querySelector(this.closeSelector);
```
levimcg: The more I think about this it probably makes more sense to set the dismiss button as instance property in the constructor that way we don't need to query for it again in the `destroy()` method.

So in the constructor you could set up the property that will hold the reference:

```js
constructor() {
// other existing properties
this.closeButton = null;
}
```

this in the `init()` method:
```js
this.closeButton = this.element.querySelector(this.closeSelector);
this.closeButton.addEventListener....
```

then in the `destroy()` method you wont need to do that set up again.

```js
// destroy()
this.closeButton.addEventListener('click', this._handleClick, false);
```

PR #238, opened 2019-12-04
Refactored file input component

zttodd:
This one isn't quite ready to be merged in, but I would still like some initial feedback on it. I'm mainly looking for feedback about:
- where to place the custom events - I can see value for developers to have a custom event fire _before_ and _after_ the file upload change even happens, since there will be different data for each. So, should it be placed at the beginning or end of the "build" methods, or maybe both?
- whether the `_handleChange` method needs to be broken down further to other methods for better readability and organization
- how we can best provide the ability to completely disabling the file input event using a custom event, similar to `event.preventDefault()` in other components. Without any JavaScript, the browser's file selection modal still appears when the matching input/label is clicked

- [x] Component should be an ES6 `class`
- [x] Component should be able to be programmatically controlled via "public" methods
- [x] Components should use data attributes as hooks for functionality and follow the format `data-[component-name]`
- [x] Should be tested and function all evergreen modern browsers and IE11
- [x] All methods should be documented using [JSDoc-style comments](https://jsdoc.app/index.html)

### DIFF
--- a/src/components/file-input/file-input.config.yml
+++ b/src/components/file-input/file-input.config.yml
@@ -3,14 +3,14 @@ label: File Input
 status: "ready"
 collated: true
 context:
-    id: "my-file-input"
+    id: "singleFileInput"
     buttonText: Upload a file
     previewText: No file selected
     fileDescription: "file-description"
 variants:
   - name: "secondary"
     context:
-        id: "my-file-input-two"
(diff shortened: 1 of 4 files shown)


zttodd:
@levimcg after making a few updates to selectors and data-attribute names, this one is ready for a more thorough review.

PR #244, opened 2019-12-16
Refactored modal js

illusivesunrae:
Refactored modal to reflect new js structure.

To Do:
- [x] Write Cypress tests
- [ ] Test browser compatibility
- [x] Fix merge conflicts


illusivesunrae:
Alright, so I started browser testing, but am running into issues in safari.

On Safari for mac I am running into issues with the focus for tabbing.

On Safari for windows, the page won't even load.

Otherwise, no issues I can find.

src/js/components/modal.js
```diff
@@ -0,0 +1,303 @@
...
+      'modalClose',
+      this.element,
+      {
+        id: this.element.dataset.modal
+      }
+    );
+
+    if (!closeEvent) return;
+
+    this.element.setAttribute('aria-hidden', 'true');
```
levimcg: Generally speaking, we should be using the native `hidden` attribute moving forward to show hide elements. Let's update that here to match how we're writing the rest of the components. There will probably be a few other placed elsewhere in the modal where you'll need to update that.

src/js/components/modal.js
```diff
@@ -0,0 +1,303 @@
...
+    if (!this.element) {
+      throw new Error(`Could not find a modal with the value of ${this.modalDataValue}`);
+    }
+
+    this.element.focus();
+  }
+
+  init() {
+    if (!this.openOnInit) {
+      this.element.setAttribute('aria-hidden', 'true');
```
levimcg: See previous comment on using `hidden` over `aria-hidden`.

Issue #136, opened 2019-07-18
Dropdown arrow keys aren't being captured causing page to jump


jolamar:
https://rivet.iu.edu/components/navigation/dropdown/ As an example open a dropdown and use the down arrow and notice the page jumps.

jolamar:
This seems to be related to event.preventDefault in this commit (which was removed):
https://github.com/indiana-university/rivet-source/commit/7c50cf373eeabe99a9a0efe367eaf143f24fcac1#diff-c21c78863bc58f6e03837c5c7788f148

zttodd:
Released with PR #294.

PR #248, opened 2020-01-07
Refactored dropdown JS

zttodd:
**In this PR:**
- Refactored component JS to ES6
- Set eventListener to the `document` so multiple instances of the dropdown can be managed within a page
- Close all other dropdowns when a new one is opened
- Added option to enable/disable "prevent page scroll" when the up and down arrows are pressed

**Development checklist:**
- [x] Component should be an ES6 `class`
- [x] Component should be able to be programmatically controlled via "public" methods
- [ ] Component should emit `customEvent`s as hooks when appropriate
- [ ] Public component methods should have callback functions where appropriate (let's talk more about this as we get started)
- [x] Components should use data attributes as hooks for functionality and follow the format `data-[component-name]`
- [ ] `customEvent`s should be cancelable by the developer using them
- [x] Should be tested and function all evergreen modern browsers and IE11
- [x] All methods should be documented using [JSDoc-style comments](https://jsdoc.app/index.html)
- [x] Component should have end-to-end (Cypress) tests that cover the functionality of all public methods, interactions e.g. open/close, show/hide, and any `customEvent`s the component emits.
- [x] Prevent page from scrolling/jumping when the arrow keys are used #136

### DIFF
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -6,6 +6,12 @@
 export default class Dropdown {
   constructor(element) {
     this.element = element;
+    this.focusableElements = 'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="0"]';
+    this.toggleAttribute = '[data-dropdown-toggle]';
+    this.menuSelector = '[data-dropdown-menu]';
+
+    // Bind methods
+    this._handleClick = this._handleClick.bind(this);
 
     this.init();
   }
@@ -35,10 +41,10 @@ export default class Dropdown {
   }
 
   init() {
-
+    this.element.addEventListener('click', this._handleClick, false);
   }
 
   destroy() {
-
+    this.element.removeEventListener('click', this._handleClick, false);
   }
 }
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -1,0 +1,44 @@
+/**
+ * Copyright (C) 2018 The Trustees of Indiana University
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+export default class Dropdown {
+  constructor(element) {
+    this.element = element;
+
+    this.init();
+  }
+
+  openMenu() {
+
+  }
+
+  closeMenu() {
+
+  }
+
+  toggle() {
+
+  }
+
+  _setUpMenu() {
+
+  }
+
+  _handleClick() {
+
+  }
+
+  _handleKeydown() {
+
+  }
+
+  init() {
+
+  }
+
+  destroy() {
+
+  }
+}
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -18,14 +18,12 @@ export default class Dropdown {
     this.menuAttribute = '[data-dropdown-menu]';
     this.menuElement = this.element.querySelector(this.menuAttribute);
 
-    // Keyboard Codes
-    this.keys = keyCodes;
-
     // Keeps track of the currently active dropdown and helps with focus management
     this.activeDropdown = null;
 
     // Bind methods
     this._handleClick = this._handleClick.bind(this);
+    this._handleKeydown = this._handleKeydown.bind(this);
 
     this.init();
   }
@@ -64,8 +62,24 @@ export default class Dropdown {
     menuList.setAttribute('aria-hidden', 'true');
   }
 
-  _setUpMenu() {
+  _setUpMenu(menu) {
+    const menuObject = {};
 
+    // Create a real Array of all the focusable elements in the menu
+    const menuFocusables = Array.prototype.slice.call(
+      menu.querySelectorAll(this.focusableElements)
+    );
+
+    // Create a property to hold an array of all focusables
+    menuObject.all = menuFocusables;
+
+    // Create a property with a reference to the first focusable
+    menuObject.first = menuFocusables[0];
+
+    // Create a property with a reference to the last focusable
+    menuObject.last = menuFocusables[menuFocusables.length - 1];
+
+    return menuObject;
   }
 
   _handleClick(event) {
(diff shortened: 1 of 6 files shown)


src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+    this.init();
+  }
+
+  /**
+   * Opens a dropdown menu
+   * @param {HTMLButtonElement} toggle - a button element which toggles the dropdown
+   * @param {HTMLDivElement} menu - a div containing the dropdown menu items
+   */
+  openMenu(toggle, menu) {
+    const toggleButton = toggle || this.toggleElement;
```
levimcg: Related to the previous comment. you can remove these variable reassignments since you can use the instance properties that hold reference to the toggle and menu instead.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+    const menuList = menu || this.menuElement;
+
+    // Return if disabled dropdown is being opened programmatically
+    if (toggleButton.hasAttribute('disabled')) {
+      return;
+    }
+
+    // If the menu was opened by clicking an associated toggle
+    if (toggleButton && toggleButton !== null) {
```
levimcg: You shouldn't need this logic here because right now it will alway return true. We should try to keep these open/close, show/hide methods as logic-free as possible.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+      toggleButton.setAttribute('aria-expanded', 'true');
+    }
+
+    // Remove the 'hidden' attribute to show the menu
+    menuList.setAttribute('aria-hidden', 'false');
```
levimcg: Let's use the `hidden` attribute instead of `aria-hidden` here since we have been moving that direction in all of the other components. So for example, this would become:

```js
this.menuElement.removeAttribute('hidden');
```

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+
+    this.activeDropdown = this.element;
+  }
+
+  /**
+   * Closes a dropdown menu
+   * @param {HTMLButtonElement} toggle - a button element which toggles the dropdown
+   * @param {HTMLDivElement} menu - a div containing the dropdown menu items
+   */
+  closeMenu(toggle, menu) {
```
levimcg: See previous comments about the `openMenu` method above. Remove variable reassignments in favor of instance properties.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+    const toggleButton = toggle || this.toggleElement;
+    const menuList = menu || this.menuElement;
+
+    // Return if disabled dropdown is being closed programmatically
+    if (toggleButton.hasAttribute('disabled')) {
+      return;
+    }
+
+    if (toggleButton && toggleButton !== undefined) {
```
levimcg: Same as previous comment about the `openMenu` method. You can remove this logic.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+      toggleButton.setAttribute('aria-expanded', 'false');
+    }
+
+    menuList.setAttribute('aria-hidden', 'true');
```
levimcg: Same as comment above, use `hidden` attribute instead of `aria-hidden` for showing/hiding of the menu.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+  _handleClick(event) {
+    const dropdown = event.target.closest(this.dropdownAttribute);
+
+    // If click didn't come from within dropdown component, close all open menus
+    if (!dropdown) {
+      this.closeMenu();
+      return;
+    }
+
+    const toggleButton = dropdown.querySelector(this.toggleAttribute);
```
levimcg: Same as comments above, you shouldn't need to reassign these here. You can reference the instance properties for toggle and menu instead.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+    const menuList = dropdown.querySelector(this.menuAttribute);
+
+    // Delegate event to only this instance of the dropdown
+    if (dropdown === this.element) {
```
levimcg: Let's look at cleaning up this logic a bit. There are a lot of nested if statements here which is generally a sign that some of this logic could be extracted/refactored. I would start by thinking about how you can pull out/combine some of these if statements to know for sure that the click didn't come from within the menu and wasn't on the toggle.

For example you could confirm that the click came from within the menu by doing something like this:

```js
if (this.menuElement.contains(event.target)) return;
/**
 *  After this you would know that the click came from within the menu
 * so you can stop here because you don't want to open or close the menu
*/
```

I'm happy to do some pair refactoring on this if you want to schedule some time to get together. Just let me know.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+      this.closeMenu();
+    }
+  }
+
+  _handleKeydown(event) {
+    const dropdown = event.target.closest(this.dropdownAttribute);
+
+    if (!dropdown) return;
+
+    const toggleButton = dropdown.querySelector(this.toggleAttribute);
```
levimcg: Same as above with these reassignments. You can reference `this.toggleElement`, etc. instead.

src/js/components/dropdown.js
```diff
@@ -0,0 +1,286 @@
...
+          this.activeDropdown = null;
+
+          break;
+        }
+  
+        case keyCodes.tab: {
+          /**
+           * Handle tab key when inside the open menu.
+           */
+          if (event.target.closest(this.menuAttribute) !== null) {
```
levimcg: I'm seeing a console error here in IE11. Not super urgent right now, but we'll need to look into this and figure out how to stop that from happening. It doesn't break anything, but we'll need to get if figured out eventually. 👍

zttodd:
@levimcg the items above have been refactored and are ready for another review. Thanks again for the feedback and the peer refactoring on `_handleClick`! 👍

src/js/components/dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+    this.activeDropdown = this.element;
+  }
+
+  close() {
+    this.isOpen = false;
+
+    this.toggleElement.setAttribute('aria-expanded', 'false');
+
+    this.menuElement.setAttribute('hidden', '');
+  }
```
levimcg: I wonder if we should we should set `this.activeDropdown = null` here when the dropdown get's closed. I'm not sure on that. 🤔

src/js/components/dropdown.js
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Creates an object of menu items within a dropdown
+   * @param {HTMLDivElement} menu - a div containing the dropdown menu items
+   * @returns {Object} An object of menu item anchor elements
+   */
+  _setUpMenu(menu) {
+    const menuObject = {};
+
+    // Create a real Array of all the focusable elements in the menu
+    const menuFocusables = Array.prototype.slice.call(
```
levimcg: There is a utility function in `js/utilities/domHelpers.js` called `nodeListToArray` that you could import and use here. It's literally this same code under the hood. 👍

src/js/components/dropdown.js
zttodd: `this.activeDropdown` is set back to `null` within the `escape` key logic, so I think this would be a good idea!

src/js/components/dropdown.js
zttodd: Thanks for the heads up! Including that now.

PR #251, opened 2020-01-08
Refactored tabs js

illusivesunrae:
Refactored tabs to reflect new js structure.

To Do:
- [x] Write Cypress tests
- [x] Test browser compatibility
- [x] ~~Merge changes from `2.0.0-dev`~~
- [x] Refactor additional tab examples in fractal


src/js/components/tabs.js
```diff
@@ -0,0 +1,220 @@
...
+        id: tab.dataset.tabPanel,
+      }
+    );
+
+    if (!activationEvent) return;
+
+    const trigger = this.element.querySelector(`[${this.tabAttribute}="${tab.dataset.tabPanel}"]`);
+
+    this.panels.forEach(panel => {
+      if (panel.getAttribute(this.panelAttribute) !== tab.dataset.tabPanel && !panel.hasAttribute('hidden')) {
```
levimcg: The addition `&& !panel.hasAttribute('hiden')` check shouldn't be needed here. All you need to know is that the current iteration (panel) _is not_ equal to the tab panel that was passed in. Since only one panel should ever be visible at a time the additional check is redundant.

src/js/components/tabs.js
```diff
@@ -0,0 +1,220 @@
...
+        this.deactivateTab(panel);
+      }
+    });
+
+    // Activate the appropriate tab/panel pair
+    tab.removeAttribute('hidden');
+    tab.classList.add('rvt-tab-open');
```
levimcg: As far as I can tell this `rvt-tab-open` CSS class  isn't used for anything and it didn't exist in the old version. I'd say we can get rid of that.

src/js/components/tabs.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  /**
+   * This function is used to trigger the 'tabDeactivated' custom event,
+   * and deactivates a tab/panel pair. It adds the hidden attribute, removes
+   * the .rvt-tab-open class, and sets the tab's aria-selected attribute to
+   * false. It also allows developers to setup a custom callback function.
+   * @param {Function} callback 
+   */
+  deactivateTab(tab, callback) {
```
levimcg: This method only ever gets called once inside the `activateTab()` method, which feels a little bit odd to me. I think we could get away with removing the important parts of this into the logic inside of `activateTab` and simplify things a bit.

src/js/components/tabs.js
```diff
@@ -0,0 +1,220 @@
...
+    // Trigger tabDeactivated custom event. This event is used to control the process of switching between tabs.
+
+    const deactivationEvent = dispatchCustomEvent(
```
levimcg: For example, I don't feel like we need a `tabDeactivated` event in addition to a `tabActivated` event. If you look at where you're calling `this.deactivateTab` in the `activateTab` method there's really nothing functionally going on other than querying the DOM for the toggle element.

Also, the way the `activateTab` method is set up right now, the `tabDeactivated` Event will be fired for every tab except the one that get activated. So in the example `tab.njk` it gets fire three times in a row. 

I'd say for now let's just stick with `tabActivated` for the custom events.

Without this custom event logic I'd say we should definitely get rid of the this `deactivateTab` method.

src/js/components/tabs.js
```diff
@@ -0,0 +1,220 @@
...
+    trigger.setAttribute('aria-selected', 'false');
+    trigger.setAttribute('tabindex', '-1');
+
+    if (callback && typeof callback === 'function') {
+      callback();
+    }
+  }
+
+  init() {
+    if (this.openOnInit) { this.activateTab(this.openOnInit); }
```
levimcg: See previous comments on setting `this.openOnInit` during set up in the constructor. With that in mind you won't need this check anymore because `this.openOnInit` will either be the tab that was initialized via a data attribute in the DOM, or the first tab. So this can become:

```js
this.activateTab(this.openOnInit);
```

src/components/tab/tab.njk
```diff
@@ -11,28 +11,25 @@
...
                 aria-selected="false"
                 data-tab="tab-4"
-                id="t-four"
-                tabindex="-1">
+                id="t-four">
             Tab four
         </button>
     </div>
-    <div class="rvt-tabs__panel" tabindex="0" role="tabpanel" id="tab-1" aria-labelledby="t-one">
+    <div class="rvt-tabs__panel" tabindex="0" role="tabpanel" id="tab-1" aria-labelledby="t-one" data-tab-panel="tab-1" hidden="">
```
levimcg: Just a FYI, you don't need to set the `hidden` attribute to an empty string here. It can be a `hidden` boolean attribute. You only have to set it to an empty string when setting `hidden` as a property in JavaScript.

src/js/components/tabs.js
```diff
@@ -0,0 +1,186 @@
...
+    // If not a tab, ignore
+    if (!currentTab) return;
+
+    // Create an array of all the focusable elements within the modal
+    const nextTab = this.tabs.indexOf(currentTab) + 1;
+    const prevTab = this.tabs.indexOf(currentTab) - 1;
+
+    switch (event.keyCode) {
+      case keyCodes.right || keyCodes.down: {
+
```
levimcg: Since there is only one statement (plus a `break`) in this case statements let's go ahead and close up the empty lines, and instead and line breaks between each `case {}` block. Right now there are no empty spaces between.

So as an example these first two would become:

```js
case keyCodes.right || keyCodes.down: {
  !this.tabs[nextTab] ? this.tabs[0].focus() : this.tabs[nextTab].focus();
  break;
}

case keyCodes.down: {
  !this.tabs[nextTab] ? this.tabs[0].focus() : this.tabs[nextTab].focus();
  break;
}
```

src/js/components/tabs.js
```diff
@@ -0,0 +1,186 @@
...
+    // Add click handlers
+    this.tablist.addEventListener('click', this._handleClick, false);
+    this.tablist.addEventListener('keydown', this._handleKeydown, false);
+  }
+
+  destroy() {
+    this.tablist.removeEventListener('click', this._handleClick, false);
+    this.tablist.removeEventListener('keydown', this._handleKeydown, false);
+  }
+
```
levimcg: Same as comment above, let's close up empty lines at the closing of nested blocks.

PR #258, opened 2020-01-16
Added first draft of documentation for Sidenav

zttodd:
In this PR is a first draft of the docs for the Sidenav. Not currently included is the accessibility section, since we will need to evaluate how to best approach keyboard navigation for this component. Once that has been figured out, accessibility notes and implementation will be added to these docs.

This is my first doc for a component, so any and all feedback would be great! I mainly used the Menu component's docs as a baseline for comparison. Since the markup requirements are fairly rigid for this component to work correctly, I didn't include many notes about that, so feedback in that area would be great if necessary.

### DIFF
--- a/src/components/sidenav/README.md
+++ b/src/components/sidenav/README.md
@@ -1,0 +1,1 @@
+# Sidenav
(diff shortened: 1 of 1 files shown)


zttodd:
Definitely! I'll update this PR with the new docs once they're ready for review.

zttodd:
Good call! I think _menu_ is more descriptive than _list_.

Issue #249, opened 2020-01-07
Vanilla JS Components: Create new Accordion component


levimcg:
## Summary
Rivet doesn't have an official Accordion component as it were. The [Collapsible add-on](https://rivet.iu.edu/add-ons/rivet-collapsible/) has to option to be configured as a component, but for Rivet 2 I'd like to see the Accordion become its own component and then move the Collapsible add-on into Rivet core as a separate component (separate issue) that provides as a progressive disclosure/show/hide functionality only (similar to the native HTML `details` element).

## Describe the solution you'd like
The existing code for the Collapsible accordion functionality should be a good start in terms of building the toggle functionality and support for keyboard navigation, but will need to be adapted to match the JavaScript design patterns we're currently using for the other Vanilla JS Components.

As far as CSS/styling goes, we can use the existing Collapsible styles for now and revisit the design a little further down the line as the 2.0 visual style starts to firm up a little more.

## Similar features or implementations
- [Existing Collapsible accordion configuration](https://rivet.iu.edu/add-ons/rivet-collapsible/)
- [Web Style Guid Accordion Chunk](https://styleguide.iu.edu/pattern-library/chunks/accordion.html)

## Additional context
Here's a link to the ARIA Authoring requirements for states, roles, keyboard functionality, etc. https://w3c.github.io/aria-practices/#accordion

### Note on keyboard functionality
Although the ARIA spec marks some keyboard functionality as option I'd like to fully implement the recommendation. That means the following:

- Focus should cycle from the last panel header to the first and vice versa when the down/up arrow key is pressed.
- Focus should move to the last panel header when the `End` key is pressed
- Focus should move to the first panel header when the `Home` key is pressed
- See the [_Keyboard Interaction_ section](https://w3c.github.io/aria-practices/#keyb) in the ARIA docs for details

illusivesunrae:
Released with PR #294.

PR #260, opened 2020-01-16
Refactored accordion js

illusivesunrae:
Refactored rivet-collapsible as a singular accordion component reflecting the new js structure.

To do:
- [ ] Test browser compatibility

For discussion:
The [WAI-Aria authoring practices](https://w3c.github.io/aria-practices/#accordion) mentions `home` and `end` keys as optional. This seems like an easy item to include. Thoughts?


levimcg:
I think it'd be pretty straight-forward to go ahead an implement the optional ARIA recommendations at this point since we're already tracking all the first and last toggle elements.

See the _Note on keyboard functionality_ [here in the original issue](https://github.com/indiana-university/rivet-source/issues/249) write up for more info. 😜

src/js/components/accordion.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  /**
+   * This function is used to initialize accordions on load. It also adds the
+   * appropriate aria-expanded, and hidden attributes to remove some burden
+   * from the develop.
+   */
+  _openOnInit() {
+    // If accordion is set to open all panels on init, open them
+    if (this.element.hasAttribute('data-accordion-init')) {
```
levimcg: In the side nav we are offering an option `openAllOnItinit` options as part of the second argument to the component construction (options object). I wonder if it make sense to do that here instead of using a data attribute? Either way the API for how we off this kind of "open all on init" functionality should be consistent. either attributes for properties on the config object (second argument to the constructor).

src/js/components/accordion.js
```diff
@@ -0,0 +1,270 @@
...
+      // Otherwise, keep all accordions closed to start
+      this.panels.forEach((panel, index) => {
+        // Set all of the accordion panels to hidden
+        this.panels[index].setAttribute('hidden', '');
+        let trigger = this.element.querySelector(`[${this.triggerAttribute}="${panel.dataset.accordionPanel}"]`);
+        // Set 'aria-expanded' for all of the accordion triggers to 'false'
+        trigger.setAttribute('aria-expanded', 'false');
+      });
+    }
+  }
```
levimcg: These `if/else` blocks are all duplication except for one check for the `openOnInit` option/panel. That most likely means these could be refactored into one `forEach` loop where you can do your logic.

src/js/components/accordion.js
```diff
@@ -0,0 +1,270 @@
...
+  /**
+  * This function is used to determine whether an allowable number of accordion
+  * panels has been set to open on initialization.
+  * @param {Node} entity - the Node in which the given selector should be found
+  * @param {string} selector - an element selector
+  * @param {Function} callback - a function the user wishes to trigger if the
+  * selector is below the maximum allowed number
+  * @param {number} max - the maximum number of a given selector allowed
+  */
+  _quantifySelector(entity, selector, callback, max = 1) {
```
levimcg: I'm 50/50 on this method/helper. It adds some additional mental overhead. I had to keep jumping back and forth between this definition site and where it's used in the constructor to understand what was happening. If we're doing these kinds of checks in every component, maybe we could refactor it out into a util/helper function and just `import` it where needed? That's said it may be a bit of a premature abstraction when you could just as easily run a check for too many attributes/initialized panels while you're doing all the other set up in the constructor.

src/js/components/accordion.js
```diff
@@ -0,0 +1,270 @@
...
+    this.destroy();
+
+    // Add click handlers
+    document.addEventListener('click', this._handleClick, false);
+    document.addEventListener('keydown', this._handleKeydown, false);
+  }
+
+  destroy() {
+    document.removeEventListener('click', this._handleClick, false);
+    document.removeEventListener('keydown', this._handleKeydown, false);
```
levimcg: These event listeners should be called on `this.element` instead of `document`. Right now if you add more than one accordion to the page (which will be a likely scenario) the event handlers break.

src/js/components/accordion.js
```diff
@@ -20,35 +30,51 @@ export default class Accordion {
...
-          }
-        });
+      this.element.querySelectorAll(`[${'data-accordion-panel-init'}]`).length >
+      1
+        ? (excess = true)
+        : (excess = false);
+
+      if (excess === true) {
+        throw new TypeError('Caught');
+      }
```
levimcg: This conditional variable setting using the ternary based on the number of elements with the `data-accordion-panel-init` could be simplified a lot by querying in `this.element`, storing that in a variable, and then checking that in your `if` statement. You could eliminate the ternary all together.

```js
const initializedPanels = this.element.querySelectorAll(
  `[${"data-accordion-panel-init"}]`
);

if (initializedPanels.length > 1) {
  throw new TypeError('Caught');
}
```

levimcg: Also, just and FYI, if you're not interpolating any data into this template literal (backticks) statement, you don't need to use the `${}` syntax here to wrap the `data-accordion-panel-init` attribute. 

So it should be the regular attribute selector:
```js
const initializedPanels = this.element.querySelectorAll(
  `["data-accordion-panel-init"]`
);
```

src/js/components/accordion.js
```diff
@@ -259,12 +264,12 @@ export default class Accordion {
     this.destroy();
```
levimcg: I'm pretty sure you can remove this call to `this.destroy()`. In the old version of the collapsible/accordion this made sense because there was one global instance instead of individual instances. In  this context (inside the `.init()` method) calling `this.destroy()` would have any affect because the instance you are creating when this code runs doesn't exist yet.

PR #262, opened 2020-01-17
Added logic for folds expanded in markup

zttodd:
In this PR:

- The `addEventListener` was moved to the beginning of the `init()` so it's run before any `return`s are called
- Logic for `this.openAllOnInit` was moved up front so the rest of the code can be skipped if it's enabled
- Added logic to handle toggle/menu pairs whenever a toggle is set to `aria-expanded="true"` in the markup before any JS is run. Any toggle that has `aria-expanded="true"` added in the markup will initialize opened on page load, and the rest will remain closed.

It would be good to have some feedback on this line in `sidenav.js`–

`childMenus.forEach(childMenu => childMenu.setAttribute('hidden', ''));`

–since this is closing all of the childMenus, and then later on some might be getting re-opened if the toggle is set to open within the markup.

### DIFF
--- a/src/js/components/sidenav.js
+++ b/src/js/components/sidenav.js
@@ -92,14 +92,6 @@ export default class Sidenav {
     // Add click handler
     this.element.addEventListener('click', this._handleClick, false);
 
-    // Expand all if openAllOnInit option is set
-    if (this.openAllOnInit) {
-      menuToggles.forEach(menuToggle => menuToggle.setAttribute('aria-expanded', 'true'));
-      childMenus.forEach(childMenu => childMenu.removeAttribute('hidden', ''));
-
-      return;
-    }
-
     // Get all the necessary DOM elements and convert to Arrays.
     const menuToggles = nodeListToArray(
       this.element.querySelectorAll(this.toggleSelector)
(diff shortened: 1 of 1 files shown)


src/js/components/sidenav.js
```diff
@@ -89,31 +89,43 @@ export default class Sidenav {
   }
   
   init() {
+    // Add click handler
+    this.element.addEventListener('click', this._handleClick, false);
+
+    // Expand all if openAllOnInit option is set
+    if (this.openAllOnInit) {
```
levimcg: You'll need to move the `const menuToggles` and `const childMenus` declarations below up above this `if` check. Right now if someone passes in the `openAllOnInit` option as `true` those declarations will be unreachable, so when you do `menuToggles.forEach` it will error out.

src/js/components/sidenav.js
```diff
@@ -89,31 +89,43 @@ export default class Sidenav {
...
     );
     const childMenus = nodeListToArray(
       this.element.querySelectorAll(this.listSelector)
     );
-    
+
+    // Hide all child menus
+    childMenus.forEach(childMenu => childMenu.setAttribute('hidden', ''));
+
     menuToggles.forEach(menuToggle => {
```
levimcg: Something to think about here. You're calling `.forEach` four different times here. Given that you know a menu toggle and a child menus _should_ always have a set of matching data attribute values (`data-sidenav-toggle`/`data-sidenav-list`). You might not need to loop through the `childMenus` and `menuToggles` arrays multiple times each. You might think about how you can use some logic inside of of each `forEach` iteration. We can spend some time pairing on this if you want.

zttodd: This logic has been refactored. Have another look and let me know your thoughts!

PR #263, opened 2020-01-21
Changed text-indent to padding-left

zttodd:
Fixes issue with wrapping text not indenting properly.

Screen Shot 2020-01-21 at 9 53 55 AM

### DIFF
--- a/src/sass/components/_sidenav.scss
+++ b/src/sass/components/_sidenav.scss
@@ -93,9 +93,9 @@ $shade-selector: '.#{$item} > .#{$list}';
 
 // Set up level indent/background shade map
 $level-styles: (
-  20: $color-black-000,
-  40: $color-black-100,
-  60: $color-black-200
+  34: $color-black-000,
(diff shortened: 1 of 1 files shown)


src/sass/components/_sidenav.scss
```diff
@@ -109,7 +109,7 @@ $level-styles: (
   // Indent the link text
   #{$indent-selector} > .rvt-sidenav__item-wrapper > .rvt-sidenav__link,
   #{$indent-selector} > .rvt-sidenav__link {
-    text-indent: $indent + px;
+    padding-left: $indent + px;
```
levimcg: Since we use `rem`s for units everywhere else we should probably convert these pixel values to rems as well. Not critical for this PR, but maybe at least add a note/comment to come back and fix that.

zttodd: I went ahead and updated the values from `px` to `rem` since it was a small amount of work to fix it. Let me know what you think about the changes!

PR #271, opened 2020-01-27
Removed normalize.css

zttodd:
In this PR:

`libs/normalize` in `rivet.scss` has been commented out to disable all `normalize.css` styles. To test out the differences with or without normalize.css, this line can be uncommented and commented to enable and disable. Once everything looks good, this line and the `libs/_normalize.scss` file can be removed entirely.

- Set `font` on all elements to `inherit` (`_base.scss`)
- Set `button` margins to 0 to override Safari user agent stylesheet defaults (`_base.scss`)
- Added `-moz-focus-inner` to remove border on focused buttons within Firefox (`_base.scss`)
- Removed scrollbar from `textarea` in IE (`_inputs.scss`)
- Set `border-collapse` and `border-spacing` to match normalize (`_tables.scss`)

### DIFF
--- a/src/js/components/accordion.js
+++ b/src/js/components/accordion.js
@@ -50,7 +50,6 @@ export default class Accordion {
 
     // Set this.openOnInit if needed
     try {
-
       if (initializedPanels.length > 1) {
         throw new TypeError('Caught');
       }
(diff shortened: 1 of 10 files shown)


zttodd:
@levimcg Good call on adding the `cursor: pointer`! I've updated the components you noted, including the `dropdown` one as well. It makes sense to make sure a cursor is applied, in case it's not set up as a `button`. But we can revert this later if needed, or now if you think it's not necessary in this PR.

Issue #265, opened 2020-01-22
Implement RFC 0002 - Container and grid simplification


levimcg:
Creating this issue to prioritize work no the _Container and grid simplification_ RFC. I'll be leaving the RFC open for a while as we work out the details on a more standard timeline for RFCs and to see if we get anymore feedback in the coming weeks, but we can go ahead and get started on this.

See [Detailed design section](https://github.com/indiana-university/rivet-rfcs/blob/rfc/container-and-grid-simplification/text/0002-container-and-grid-simplification.md#detailed-design) of the RFC text for implementation details.

illusivesunrae:
Released with PR #294.

PR #273, opened 2020-01-28
Refactored grid

illusivesunrae:
In this PR:

- Renamed class `rvt-grid` -> `rvt-row`
- Renamed class `rvt-grid_item` -> `rvt-cols`
- Refactored containers to center by default
- Refactored container sizes to standard `sm`, `md`, `lg`, and `xl`
- Refactored containers to require a single class instead of 3 
(ex. `rvt-container rvt-container--sophomore rvt-container--center` -> `rvt-container-md`)
- Added gutter options for `rvt-row--loose` and `rvt-row--tight` classes
- Refactored responsive class variants from `#-bp-up` to `4-bp`

See [issue](https://github.com/indiana-university/rivet-source/issues/265) based on [RFC](https://github.com/indiana-university/rivet-rfcs/blob/rfc/container-and-grid-simplification/text/0002-container-and-grid-simplification.md#detailed-design) for more details.

___

There is still more refactoring of the code that can be done. Would love feedback on that especially.


levimcg:
Thanks for digging into this one! Just a heads-up I'm looking at this PR, but want to spend a little more time with it. Might be tomorrow before I get a review finished. 👍

src/sass/components/_grid.scss
```diff
@@ -88,7 +86,7 @@ $gutter: $md;
   }
 
   &--tight {
-    margin: 0 ($gutter/-2);
+    margin: 0 ($gutter/-1);
   }
```
levimcg: Right now this `--tight` modifier has no affect since you're dividing by negative on (`-1`). You'll need to divide by something less than one to make the gutters smaller which is the intended affect of the `--tight` row modifier. Maybe `.75` to start with?

src/sass/components/_grid.scss
```diff
@@ -112,7 +110,7 @@ $gutter: $md;
...
 
   &--tight > [class^="#{$prefix}-cols"] {
-    padding: 0 ($gutter/2);
+    padding: 0 ($gutter);
   }
```
levimcg: See previous comment. This will need to be the positive version of the negative number you'll need to divide by in the `rvt-row--tight` modifier above.

src/sass/components/_grid.scss
```diff
@@ -1,170 +1,174 @@
...
     min-width: 0;
   }
+
+  &--loose > [class^="#{$prefix}-cols"] {
+    padding: 0 ($gutter * 1.66);
+  }
+
+  &--tight > [class^="#{$prefix}-cols"] {
+    padding: 0 ($gutter * .75);
+  }
```
levimcg: This is kind of nit-picky, but can you move these two modifier rules up so they come right after the original `--tight` and `--loose` declarations? It's a little easier to understand what the modifiers are doing if you can see the styles they're applying in close proximity to each other.

PR #283, opened 2020-02-07
Created design tokens

zttodd:
In this PR:

- Installed Style Dictionary
- Reconstructed Sass variables and maps as design tokens (`src/tokens/`)
- Created `.tokens.config.js` and adjusted `gulpfile.js` to compile design tokens into `src/sass/core/` for consumption by Rivet
- Segmented filters, formats, and file output configs in `src/tokens/`

For more detailed information about how tokens are created and processed, check out the [Wiki](https://github.com/indiana-university/rivet-source/wiki/Design-tokens).


src/tokens/spacing.json
```diff
@@ -0,0 +1,36 @@
+{
+  "spacing": {
+    "base": {
+      "value": ".5rem",
+      "comment": "8px"
+    },
+    "xxs": {
+      "value": "({spacing.base.value} / 2)",
```
levimcg: Let's go ahead and do this math and use the resulting values here that way the output in various formats will be usable without having to manipulate a to do math in the target language. Seems like you can still do math in the source tokens, but it will need moved inside the the curly brackets, which I'm assuming mark the start of a JavaScript expression where you could then do the math. I'm not sure about how that works in style dictionary, so you'll need to confirm 👍 

```js
{spacing.basevalue / 2}
```

src/tokens/widths.json
```diff
@@ -0,0 +1,36 @@
...
+    "lg": {
+      "value": "({width.base.value} * 5)",
+      "comment": "320px"
+    },
+    "xl": {
+      "value": "({width.base.value} * 6)",
+      "comment": "384px"
+    },
+    "xxl": {
+      "value": "({width.base.value} * 8)",
```
levimcg: Same here, I would think about either moving the multiplication inside the curly brackets/expression (if possible), or doing the math and hard-coding the value here. The goal being that any of the compiled tokens would have the final values as strings instead of math expressions which is how they are currently being output.

zttodd:
> I just noticed a couple of things when running locally:
> 
> 1. Let's think about adding the output .scss variables and maps were using in `/core` to `.gitignore`. Since we generate those every time we start the dev process, we shouldn't need to track the compiled files in source. Kinda the same way we don't track the output CSS from `rivet.scss`. So that would be `src/sass/core/variables.scss` and `src/sass/core/maps/` (the whole folder) that gets added to .gitignore.
> 2. Looks like the class names for the `rvt-bg-*` color utilities are broken. Only the `base` colors work right now.
> 3. Also check on the width utils. It looks like they aren't working anymore either.

@levimcg – @illusivesunrae and I worked through these issues earlier, and the non-scss design token platform outputs have been removed from this PR, so this is ready for a final review!

PR #296, opened 2020-02-13
Created script for uploading release asset

zttodd:
This fixes an issue caused by `rivet.zip` file being present in our `.gitignore`, which resulted in it being unreachable by the `upload-release-asset` action. To fix this, we created a new script in `package.json` to create a renamed asset for upload to GitHub for release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "scripts": {
     "start": "gulp",
     "build": "npm run clean && gulp release && zip -r rivet.zip css js sass tokens index.html",
+    "asset:upload": "npm run clean && gulp release && zip -r rivet-gh-release.zip css js sass tokens index.html",
     "fix": "stylelint 'src/sass/**/*.scss' --fix",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
   "scripts": {
     "start": "gulp",
     "build": "npm run clean && gulp release && zip -r rivet.zip css js sass tokens index.html",
```
levimcg: Do we need to zip this stuff up in the `build` script anymore since we there's a separate step/script that being used to zip the file that gets attached to the release?

The only thing we use that .zip file for is to attach to the GitHub release, so I"m not sure the zip file creation needs to be apart of this `build` step. But, I could be wrong.

zttodd:
@illusivesunrae this fix makes sense to me, nice work! 👍 I've added it into this branch and will go ahead and merge this PR.

zttodd:
This PR fixes #297

PR #302, opened 2020-02-17
Created essay component

illusivesunrae:
In this PR:

- Created nunjucks template, config.yml for settings, and a blank README for docs
- Set up component based on this prototype:
https://codepen.io/levimcg/pen/XWrPLZK
- Fixed issue with focus border appearance
- Added opacity shift for hover


src/sass/essay/_base.scss
```diff
@@ -0,0 +1,94 @@
+// Copyright (C) 2018 The Trustees of Indiana University
```
levimcg: I believe this should be `(C) 2020` since this a new component, I believe. Let's double-check the requirements laid out for copyright banners here: https://indiana-university.github.io/

src/sass/essay/_base.scss
```diff
@@ -0,0 +1,94 @@
...
+
+a.rvt-essay {
+  &__stat:hover,
+  &__image:hover {
+    opacity: .5;
+    transition: all .3s linear;
+  }
+}
+
+@media screen and (min-width: 800px) {
```
levimcg: Let's use a variable for this value here instead of hard coding. Eg. `$breakpoint-md`. The `800px` value in the prototype wasn't chosen with any purpose, so we should pick one of our predefined breakpoints, unless there is a specific need for it to be 800px.

src/components/essay/essay.config.yml
```diff
@@ -0,0 +1,38 @@
...
+              - link: "#"
+                image: "https://www.fillmurray.com/600/400"
+            - stats:
+              - bgColor: "crimson-dark"
+                number: 39
+                caption: "Years old is old"
+              - link: "#"
+                bgColor: "crimson"
+                number: 38
+                caption: "Years old is old"
```
levimcg: Let's update this placeholder copy to something more generic/random than my stream of consciousness typing from the prototype. Also, 39 in not _that_ old. 😉

src/sass/essay/_base.scss
```diff
@@ -0,0 +1,94 @@
...
+
+a.rvt-essay {
+  &__stat:hover,
+  &__image:hover {
+    opacity: .5;
+    transition: all .3s linear;
+  }
+}
+
+@media screen and (min-width: $breakpoint-md) {
```
zttodd: Should this be refactored to `@include mq($breakpoint-md)`?

zttodd:
@levimcg do you think we might need to bump up the color contrast for the stats during a hover state? It might not be an issue with the photos, but I'm wondering if the white text behind the 50% opacity could be difficult to see for some users.

If so, one idea could be to add an `:after` pseudoelement after the `stat-number` and `stat-caption` that spans the width and height of the parent container. Something like this:
```
  &__stat:hover,
  &__image:hover {
    &::after {
      background-color: $color-crimson-base;
      content: '';
      height: 100%;
      opacity: .5;
      position: absolute;
      width: 100%;
      z-index: 0;
    }
  }
```
This would make sure that the text is always on top of the opacity overlay background, instead of the background covering up the text. We would still want to make sure it passes in the hover state.

levimcg:
@zttodd Yeah, that's a good call. I think `.5` opacity might be too light with the white on top. I like your idea with the `::after` pseudo element. 👍 

If we got that route it would makes sense to specify a specific background color for the crimson to change to on hover instead of using opacity. That way we can more easily ensure that it is meeting contrast ratio and be able to test it. By checking white against what ever that color is.

illusivesunrae:
@levimcg My only concern with switching to a specific background color rather than an opacity filter, is that the behavior won't be consistent between stat links and image links.

levimcg:
Hmm, yeah, that's a good call. Per our face-to-face conversation, mayb we can removed the images-as-links functionality for now, and take some time to think about whether or not that's a good idea.

Other than that I think this is a really good first pass at this component. It would be a good idea to add some basic documentation on how to implement the different versions with some code examples, etc. in the Fractal site.

PR #303, opened 2020-02-17
Created news component

illusivesunrae:
In this PR:

- Created nunjucks template, config.yml for settings, and a blank README for docs
- Set up component based on the the Rivet feature component and [Framework News item](https://styleguide.iu.edu/pattern-library/sections/news-events.html)

I have played around with the preferred distance between the title and the elements that follow it. I also am not sire about the placement of the date vs. the tags in comparison to the rest of the information.


levimcg:
@zttodd and I had a quick conversation about this yesterday at the end of the day. I think this kind of component could/should maybe be more generic instead of tying it to a _News_ use case. There is so much overlap between this and the _Feature_ component that we should look at how we could either a.) consolidate those two components, or b.) make this one more generic so that it could be used for news, events, a profile teaser, etc.

Let's touch base sometime today and maybe chat through this a bit.

PR #314, opened 2020-02-24
Refactored feature component

illusivesunrae:
In this PR:

- Created feature component template
- Created subdirectory of `mods` that show most potential display options
- Created subdirectory of `patterns` to show commonly used orientations for the feature component

The main template is named `feature.njk`, the mods template is named `feature-mods.njk`, and the pattern template is named `feature-pattern.njk`. This establishes a naming convention that can be reused for other components which have associated patterns.

The mods and pattern templates render the original base feature template with data from their specific configuration files.


src/sass/feature/_base.scss
```diff
@@ -4,50 +4,56 @@
...
   &__image img {
     display: block;
     width: 100%;
   }
 
   &__body {
-    margin-top: $spacing-md;
+    > * {
+      margin: 0;
+    }
```
zttodd: I talked to Levi a little about this, because the first item in each variant had a `margin-top`. To refactor this a bit, I was thinking we could do something like this:

```
    * + * {
      margin-bottom: 0;
      margin-top: $spacing-sm;
    }
```

and then remove the `margin-top` from the `&__title` and `&__detail`. That way the first item doesn't have a `margin-top`, but the following items do.

src/sass/feature/_base.scss
```diff
@@ -4,50 +4,52 @@
...
 .#{$prefix}-feature {
+  display: flex;
+  flex-direction: column;
+
+  &--horizontal {
+    flex-direction: row;
+
+    .#{$prefix}-feature__image {
+      margin-right: $spacing-sm;
+    }
```
levimcg: In the horizontal version we'll need to control the flex-shrink/grow properties here. The `__image` container shouldn't shrink as the outer container get's smaller. That will mean setting a `max-width` on the `__image` container and setting `flex-shrink: 0`. For now I think we can pick a somewhat arbitrary value for the `max-width` something around 160-180 rendered pixels will be good. So whatever that is in rem. I would make the width some multiple of a `$spacing-*` unit.

We want to avoid the `__image` container shrining like this on small screens:
image

I would also think about the way top magins are handled in the `__body` here. The top of the content should be flush with the top of the `__image` container. (see screenshot ☝️ ). I'm not sure why the first `AUTHOR` tag is getting top margin there since you are using the `*+*` selector here on the `__body` element.

src/sass/feature/_base.scss
```diff
@@ -4,50 +4,52 @@
...
+  }
+
   &__image img {
     display: block;
     width: 100%;
   }
 
   &__body {
-    margin-top: $spacing-md;
+    * + * {
```
levimcg: Something to think about here is that using the universal version of this selector instead of the direct child version Eg. `> * + *` means that _anything_ inside the `_body` element will have the top margin applied no matter how deeply nested. I think in this case we only want consistent top margin applied to the direct children (`__detail1` elements) that we know about. Using the selector as is could have unintended consequences when someone puts multiple elements inside of a `__detail` element. I don't think we should be controlling the spacing between elements that people are putting in the "slots" we are providing.

here's an example. There is top margin being applied to the _Action_ you have inside this configuration causing the display of the component to be thrown off:

image

src/components/feature/feature-mods/feature-mods.config.yml
```diff
@@ -0,0 +1,204 @@
+title: Mods
```
levimcg: This is a bit of nitpick, but I think we could use clearer language here. "Variants", "Test Combinations", or "Configuration options" might be clearer.

illusivesunrae: I think maybe we should talk through this part as a group because I'm finding these less clear. 😬

levimcg: That's fine. Let's defer to some later date when we work on the actual public facing documentation. Since this is just internal documentation for now, the conversation shouldn't stop us moving forward with the feature.

levimcg:
@illusivesunrae Oh, one more thing I meant to mention here. This has been a pretty long-running branch, so I wanted to give you a quick reminder to make sure you pull in `2.0.0-dev` before merging.

zttodd:
Aside from the changes with @levimcg's feedback, I think the docs and the refactoring all looks great. Awesome work on this!! 👍

src/sass/feature/_base.scss
```diff
@@ -4,50 +4,58 @@
 @use 'core' as *;
 
 .#{$prefix}-feature {
+  display: flex;
+  flex-direction: column;
+
+  &--horizontal {
+    flex-direction: row;
+
+    .#{$prefix}-feature__body :first-child {
```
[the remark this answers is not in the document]
illusivesunrae: So with this one, given the weird issues I was experiencing with the lobotomized owl, I needed to guarantee that on the horizontal layout, the first element within `__body` wouldn't have a `margin-top`. When I erase the the space, it targets differently than if I keep the space.

This is what the targeting description is within VS Code with the space:
Screen Shot 2020-03-05 at 10 48 25 AM

This is the description without the space:
Screen Shot 2020-03-05 at 10 48 35 AM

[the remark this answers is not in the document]
levimcg: Cool, I'm checking it out now. That's weird. I've never ran into a situation like that. 🤔 Seems like then in this case you're just targeting the pseudo element `:first-child` in general then? That's wild, I've never seen that before.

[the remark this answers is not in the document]
levimcg: Oh, I think I see why this is happening. The `> * + *` selector should take care of not applying top margin to the first element inside of `__body`, but in these examples the first element is a `p` tag for the first `__detail` element. There's a more specific rule up the cascade, if you disable that rule you can see it's the default user agent style sheet and that is somehow more specific than the lobotomized owl here. 

image

If that first `__detail` element were a generic `div` or `span` it's not a problem, but it should be tolerant enough for folks to use different elements to mark it up if needed.

The thing that threw me off was the general child select combo of the `__body :first-child`. To make it a little clearer, it would probably make sense to use the direct child and universal selector there. 

So something like:

```sass
.#{$prefix}-feature__body > *:first-child {
  margin-top: 0;
}
```

This will also make it so _only_ the first direct child gets the top margin set to 0. I could imagine a scenario where there might be elements nested inside of the a `rvt-feature__detail` element that you would want to be able to control the top margin on. the way the selector is written now, it would set the `margin-top: 0` for _any_ `:first-child` inside of _any_ `__detail` element, no matter how deeply-nested.

One more thought here. I think it would make sense to set the `__detail` element to display block by default. If I change the markup where you have "Publish date" to a `<time>` element, which would make sense in a lot of editorial contexts, the `margin-top` rule in the lobotomized owl doesn't get applied because `time` is an inline, not block element to which margins won't get applied.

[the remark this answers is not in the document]
illusivesunrae: Ahhhh. That makes so much sense! Thanks for the assist, I'll get on updating that. 🙌

PR #318, opened 2020-02-25
Added docs for avatar component

zttodd:
In this PR:

- Added docs to `README.md` within the avatar component
- Added the class `rvt-avatar__image` to the `img` within the image avatar
- Updated the avatar Sass from `img` to `rvt-avatar__image`

### DIFF
--- a/src/components/avatar/README.md
+++ b/src/components/avatar/README.md
@@ -1,0 +1,4 @@
+# Avatar
+
+A text or image-based graphic to represent a person or user
+
--- a/src/components/avatar/README.md
+++ b/src/components/avatar/README.md
@@ -2,13 +2,13 @@
 A text or image-based graphic to represent a person or user
 
 ## When to use
-Use the avatar component to visually represent a person within content or a user interface. This could include:
+Use the avatar component to visually reference a person within content or a user interface. This could include:
 
-- A photo avatar of a person who said a quote (such as when used within the `quote` component)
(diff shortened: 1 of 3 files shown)


src/components/avatar/README.md
```diff
@@ -0,0 +1,59 @@
+# Avatar
+A text or image-based graphic to represent a person or user
+
+## When to use
+Use the avatar component to visually represent a person within content or a user interface. This could include:
+
+- A photo avatar of a person who said a quote (such as when used within the `quote` component)
```
illusivesunrae: A photo of someone paired with relevant text (such as within the `quote` component)

src/components/avatar/README.md
```diff
@@ -0,0 +1,59 @@
...
+- A text or image avatar to represent a user who is logged into a system (such as when used within the `identity module` in a `header`)
```
illusivesunrae: Text or an image which represents a user within an interface (such as in the `identity module` within a `header`)

src/components/avatar/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## When to use something else
+The image avatar should be used with raster images. If a vector graphic is desired, consider using something like a SVG instead.
```
levimcg: I would think about either linking to the definition of _raster image_, or maybe changing up the words a bit. Maybe something like "...used with image formats like .jpg or .png that are sized for use on the web."

PR #321, opened 2020-02-28
Created action for automating the creation of changelogs

zttodd:
Earlier this week, @illusivesunrae and I worked on adding a GitHub action for automatically creating a changelog and adding it to the release whenever a release workflow is ran. To avoid cluttering up the commit history with trial and error experimention, we worked through the process at this repo: https://github.com/zttodd/github-js-actions

@illusivesunrae did the large majority of this PR in figuring out the logic of how everything worked, and we both peer coded through troubleshooting and debugging the action during initial testing.

Here's what we have left to do:

- Test out how this action will work within the `alpha.yml` release as a local action without publishing to the GitHub action marketplace. I talked to @levimcg about this, and because it's just an `index.js`, `action.yml`, and a couple npm packages (`@actions/core` and `@actions/github`), we should be able to point to the `index.js` from the workflow and call it in that way.
- Discuss whether the folder structure `/gh-actions/[action-package]` is the one we want to use for local GH actions that don't need to be published

### DIFF
--- a/gh-actions/build-changelog/index.js
+++ b/gh-actions/build-changelog/index.js
@@ -4,12 +4,9 @@ const { GitHub } = require('@actions/github');
 async function run() {
   try {
     const myToken = core.getInput('GITHUB_TOKEN');
-
     const tag = core.getInput('tag_name', { required: true });
     const label = tag.substr(10, tag.length - 1);
-
     const query = `type:pr+label:${label}`;
-
     const octokit = new GitHub(myToken);
 
     const { data: pullRequest } = await octokit.search.issuesAndPullRequests({
--- a/gh-actions/build-changelog/index.js
+++ b/gh-actions/build-changelog/index.js
@@ -1,0 +1,31 @@
+const core = require('@actions/core');
+const { GitHub } = require('@actions/github');
+
+async function run() {
+  try {
+    const myToken = core.getInput('GITHUB_TOKEN');
+
+    const tag = core.getInput('tag_name', { required: true });
+    const label = tag.substr(10, tag.length - 1);
+
+    const query = `type:pr+label:${label}`;
+
+    const octokit = new GitHub(myToken);
+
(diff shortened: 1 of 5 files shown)


gh-actions/build-changelog/index.js
```diff
@@ -0,0 +1,31 @@
...
+    const myToken = core.getInput('GITHUB_TOKEN');
+
+    const tag = core.getInput('tag_name', { required: true });
+    const label = tag.substr(10, tag.length - 1);
+
+    const query = `type:pr+label:${label}`;
+
+    const octokit = new GitHub(myToken);
+
+    const { data: pullRequest } = await octokit.search.issuesAndPullRequests({
```
levimcg: This github library seems pretty cool! Seems like there's lots of cools stuff we could do with it.

gh-actions/build-changelog/index.js
```diff
@@ -0,0 +1,31 @@
...
+async function run() {
+  try {
+    const myToken = core.getInput('GITHUB_TOKEN');
+
+    const tag = core.getInput('tag_name', { required: true });
+    const label = tag.substr(10, tag.length - 1);
+
+    const query = `type:pr+label:${label}`;
+
+    const octokit = new GitHub(myToken);
```
levimcg:
```suggestion
    const myToken = core.getInput('GITHUB_TOKEN');
    const tag = core.getInput('tag_name', { required: true });
    const label = tag.substr(10, tag.length - 1);
    const query = `type:pr+label:${label}`;
    const octokit = new GitHub(myToken);
```

This is more of a style observation/something I think about sometimes. Since all of these are just one-liner variable declarations, does it make more sense to have extra new lines between some of them, but not all? I think it makes sense to logically group declarations/statements, but sometimes it's just a bunch of variables you need to declare before you do the actual work later in the function body. 

Nothing right or wrong here, just thinking out loud a little bit.

PR #322, opened 2020-02-28
Added avatar variants to quote component

zttodd:
In this PR:

- Text and image `avatar` variants in different sizes were added as options for the `quote` component
- A `utility_class` option was added to the `avatar` component so a utility like `rvt-m-right-md` could be added to the `avatar` within the quote component

### DIFF
--- a/src/sass/quote/_base.scss
+++ b/src/sass/quote/_base.scss
@@ -30,6 +30,10 @@
       font-size: $ts-32 * 4;
     }
   }
+
+  .#{$prefix}-avatar {
+    margin-right: $spacing-md;
+  }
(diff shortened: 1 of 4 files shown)


src/sass/quote/_base.scss
```diff
@@ -28,6 +30,10 @@
       font-size: $ts-32 * 4;
     }
   }
+
+  .#{$prefix}-avatar {
+    margin-right: $spacing-md;
```
levimcg: Lets bump this down to `$spacing-sm`. Right now the avatar image feels too far away/disconnected from the quote text.

PR #324, opened 2020-02-28
Updated accordion styles and markup

levimcg:
This PR updates the accordion styles and markup to remove old placeholder styles and markup from the collapsible component.


levimcg:
I updated the markup on the accordion, so I'm gonna to fix the tests. They're broke right now 😬

src/sass/accordion/_base.scss
```diff
@@ -4,76 +4,49 @@
...
-    border-top: 1px solid $color-black-200;
-    padding: $spacing-sm;
-  }
-}
-
-.rvt-collapsible-accordion {
-  .rvt-collapsible:not(:first-child) {
-    margin-top: $spacing-sm;
+  &__panel {
+    padding: $spacing-md $spacing-md;
```
illusivesunrae: Wondering if this could be written simply as `padding: $spacing-md;` or if there's a specificity issue that doubling up this way handles.

zttodd:
This looks great! Just one comment – not sure if it's just me, but I think the `+ / -` might be slightly off center vertically when lined up with the text and the collapsible height.
Screen Shot 2020-03-03 at 9 57 55 AM

src/sass/accordion/_base.scss
levimcg: Oh, yeah, good call. I think they were different values at one point, but since they're the same, I'll just use the one value/shorthand.

levimcg:
Thanks for the feedback y'all. I've got some time blocked tomorrow to finish updating the tests and I'll incorporate this feedback when I do that. 👍

levimcg:
Ok, got the tests refactored and added a few more for keyboard interactions. All the tests are passing so I feel good about this being done unless anyone has anymore feedback.

PR #326, opened 2020-03-05
Created basic documentation for the action component

illusivesunrae:
In this PR:

- Created README.md for the action component and added documentation

### DIFF
--- a/src/components/action/README.md
+++ b/src/components/action/README.md
@@ -1,0 +1,48 @@
+# Action
+
+Provide visible and concise path forward to a user
+
+## When to use
+
+Use the action component when you want to spur users on to a desired endpoint. This could include:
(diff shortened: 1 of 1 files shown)


illusivesunrae:
I agree about including language that leans toward limiting usage. I think the button comparison is good, I'll work it in!

src/components/action/README.md
```diff
@@ -0,0 +1,48 @@
+# Action
+
+Provide visible and concise path forward to a user
```
levimcg: Should this be "Provide **a** visible..."? It reads a bit awkward to me the way it is.

illusivesunrae: Yup, that's totally a typo, lol.

PR #327, opened 2020-03-06
Created documentation for the grid component

illusivesunrae:
In this PR:

- created documentation for the grid
- drew from current documentation, where applicable

### DIFF
--- a/src/components/grid/README.md
+++ b/src/components/grid/README.md
@@ -1,1 +1,189 @@
 # Grid
+
+A 12 column responsive grid with support for automatic columns that makes it easy to lay out your application.
+
+## When to use
+
+Use the grid when you need to create structure within your application.
+
+## When to use something else
+
+## How it works
+
+The Rivet grid consists of containers, rows, and columns.
+
+### Containers
+
+The container is meant to be a generic wrapper to add enough padding to the content of your application so that it doesn’t bump up against the sides of the viewport. They are centered by default. The container can be used as is without any other grid items inside, but it **is required to wrap all other grid elements**.
+
+```
+<div class="rvt-container-xl">
+  <div class="rvt-row">
+    <div class="rvt-cols">
+      <span>Column</span>
+    </div>
+  </div>
+</div>
+```
+
+#### Container sizes
+
+All containers must be limited to a specific size. There are four variants that change the `max-width` of the container.
+
+- `rvt-container-sm` = 640px
+- `rvt-container-md` = 1024px
+- `rvt-container-lg` = 1140px
+- `rvt-container-xl` = 1380px
+
+### Rows
+
+The row uses `flexbox` to control the alignment and distribution of columns. **The row is required to wrap all column elements.**
+
+#### Gutters
+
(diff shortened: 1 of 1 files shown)


src/components/grid/README.md
```diff
@@ -1 +1,189 @@
...
+## When to use something else
+
+## How it works
+
+The Rivet grid consists of containers, rows, and columns.
+
+### Containers
+
+The container is meant to be a generic wrapper to add enough padding to the content of your application so that it doesn’t bump up against the sides of the viewport. They are centered by default. The container can be used as is without any other grid items inside, but it **is required to wrap all other grid elements**.
+
```
levimcg: Based on past observations of how people have used the grid, I think it would be good to add a couple notes here in the section about containers. Here are some rough notes:

## Single columns layouts
If you are not creating a column-based layout (e.g. more than one single column), you don't necessarily need to add the extra markup for rows and columns.

Example:

```html
<div class="rvt-container-md">
  <div class="rvt-row">
    <div class="rvt-cols-12">
      <p>Some single-column content...</p>
    </div>
  </div>
</div>
```

Is visually identical to:

```html
<div class="rvt-container-md">
  <p>Some single-column content...</p>
</div>
```

While it won't hurt anything, and it would logically be strange if we left out `rvt-cols-12` but created classes for cols 1-11, for developer convenience, it's easier to use the container on it's own. It might make sense to use the `rvt-cols-12` class if you're using multiple responsive grid classes to change layout at different screen sizes. Eg. `class="rvt-cols-12 rvt-cols-6-md`. That way it's clear to other developers that the item in the grid starts off full-width (12 columns) and becomes 50% (6 columns) on medium screens and larger.

## Nesting grid rows
You only need to use one outer container when nesting grids. When nesting a `rvt-row` inside a `rvt-cols-*`, you should not wrap the new `rvt-row` in another container. This will throw off the alignment of the content in nested grids.

✅ Do this:

```html
<div class="rvt-container-xl">
  <div class="rvt-row">
    <div class="rvt-cols-6-md-up">
      <p>I take up 50% of the space...</p>
      <div class="rvt-row">
        <div class="rvt-cols-6-md-up">
          <p>I'm 25% now...</p>
        </div>
        <div class="rvt-cols-6-md-up">
          <p>Me too, I'm 25%...</p>
        </div>
      </div>
    </div>
  </div>
</div>
```

🚫 Don't do this:

```html
<div class="rvt-container-xl">
  <div class="rvt-row">
    <div class="rvt-cols-6-md-up">
      <p>I take up 50% of the space...</p>
      <!-- This container isn't neccessary  and will throw off the alignement of grid content -->
      <div class="rvt-container">
        <div class="rvt-row">
          <div class="rvt-cols-6-md-up">
            <p>Uh oh, I'm 25% now, but not aligned correctly</p>
          </div>
          <div class="rvt-cols-6-md-up">
            <p>Uh oh, I'm 25% now, but not aligned correctly</p>
          </div>
        </div>
      </div>
    </div>
  </div>
</div>
```

illusivesunrae:
@levimcg I went ahead and worked the information about single column layouts into containers section, and created a subsection for the grid nesting docs.

PR #328, opened 2020-03-09
Created opt-in classes for HTML elements

zttodd:
In this PR:

Opt-in Rivet classes and example layout pages were created for the following HTML elements and headings:

- `abbr`
- `code`
- `h1`, `h2`, `h3`, `h4`, `h5`, `h6`
- `hr`
- `strong`

The Sass partials for these elements (`_elements.scss`, `_headings.scss`) were created within the `utilities` folder, but I'm up for feedback on whether there might be a better location to store these classes.


src/sass/utilities/_headings.scss
```diff
@@ -0,0 +1,20 @@
+@use 'core' as *;
+
+/**
+ * Normalize Heading sizes since we're dealing with applications not
+ * Editorial/article-like content. We'll make them slightly bigger than the
+ * base font size.
+ */
+
+.#{$prefix} {
```
levimcg: I'm not sure this is really adding much value at this point. I think we could get rid of these and just let headings that aren't set with any utility classes of specific styles (eg. custom CSS) just inhert the user-agent styles. You could accomplish these same styles with util classes, so these `rvt-h*` classes aren't super useful. Does that make sense?

I could see instead creating h1-6 styles that set `font-size` values from the type scale based on convenience. I'm not sure what the right answer is here, but I think it needs a little more thought. Maybe we can spend a bit of time chatting through this face-to-face tomorrow?

zttodd: This is a great point to explore. Let's talk more about this tomorrow 👍

zttodd: @levimcg this has been updated and is ready for another review when you have the time.

PR #329, opened 2020-03-09
Created JSON design token output

illusivesunrae:
In this PR:

- Created new format for outputting design tokens as JSON

### DIFF
--- a/src/tokens/formats/json-formats.js
+++ b/src/tokens/formats/json-formats.js
@@ -7,8 +7,7 @@ const jsonVariables = {
     let categories = Object.keys(properties);
 
     // Wrapper is the string that will be built into the final output
-    let wrapper = ``;
-    wrapper += `{\n`;
+    let wrapper = `{\n`;
 
     // Output holds the output from each category loop
     let output = ``;
--- a/src/tokens/formats/json-formats.js
+++ b/src/tokens/formats/json-formats.js
@@ -10,7 +10,7 @@ const jsonVariables = {
     let wrapper = ``;
     categories.forEach(function(item) {
       wrapper += `  ${JSON.stringify(item)}:[\n`;
-      // Get all entries with category and push to new object/array
+      let counter = 0;
       for (const key in Object.entries(allProperties)) {
         const category = allProperties[key].attributes.category;
         const propValue = allProperties[key].value;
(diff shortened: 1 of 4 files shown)


src/tokens/formats/json-formats.js
```diff
@@ -0,0 +1,70 @@
...
+  name: 'rvt/json/variables',
+  formatter: function(dictionary) {
+    const allProperties = dictionary.allProperties;
+    const properties = dictionary.properties;
+    // eslint-disable-next-line prefer-const
+    let categories = Object.keys(properties);
+
+    // Wrapper is the string that will be built into the final output
+    let wrapper = ``;
+    wrapper += `{\n`;
```
levimcg: I believe this could all be done in one go:

```suggestion
    let wrapper = `{\n`;
```

src/tokens/formats/json-formats.js
```diff
@@ -0,0 +1,70 @@
+const jsonVariables = {
+  name: 'rvt/json/variables',
+  formatter: function(dictionary) {
+    const allProperties = dictionary.allProperties;
+    const properties = dictionary.properties;
+    // eslint-disable-next-line prefer-const
+    let categories = Object.keys(properties);
```
levimcg: You should be able to tack this on the front of your big `forEach` loop below. Might be a little cleaner and less verbose.

src/tokens/formats/json-formats.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // Wrapper is the string that will be built into the final output
+    let wrapper = ``;
+    wrapper += `{\n`;
+
+    // Output holds the output from each category loop
+    let output = ``;
+    categories.forEach(function(item) {
```
levimcg: See comment above. Could be a little clearer to do the following:

```suggestion
    Object.keys(properties).forEach(item => {
```

You could alternatives eliminate the `properties` variable declaration too if you want:

```js
Object.keys(dictionary.properties).forEach(item => {
  // Do stuff with each property/dictionary entry
});
```

src/tokens/formats/json-formats.js
```diff
@@ -0,0 +1,70 @@
...
+      output += `  ${JSON.stringify(item)}:[\n`;
+
+      // Set counter to check for first item in output loop
+      let counter = 0;
```
levimcg: If I understand what's going on correctly, you may be able to get away with getting rid of this counter variable if you're only using it to check if it's the first item in the loop or not. The callback function in `forEach` accepts a second `index` argument that you could check against.

src/tokens/formats/json-formats.js
```diff
@@ -0,0 +1,70 @@
...
+           *  {
+           *      "name": "fully-articulated property name",
+           *      "value": "value"
+           *  }
+           */
+
+          let line = ``;
+
+          // If first item, do not add initial comma
+          line += counter === 1 ? `    {\n` : `,\n    {\n`;
```
levimcg: If you've passed in `index` as the second argument to you `forEach` callback, you could check that it's the first item in the array. Similar to your counter variable, if the index is `0` (first item in the array), then you don't add the comma.

```suggestion
          line += (index === 0) ? `    {\n` : `,\n    {\n`;
```

I think what you're doing here works, but since you get a counter for free with `forEach`, it might be a good opportunity to clean it up a bit, if I'm understanding things correctly. **I haven't tested this code**, but I _think_ it should work.

PR #330, opened 2020-03-09
Created calendar tile component

illusivesunrae:
In this PR:

- Created calendar tile component
- Created basic documentation


src/components/calendar/README.md
```diff
@@ -0,0 +1,27 @@
...
+## Adding the markup
+
+The container for a calendar tile is made up of a `div` with a class of `rvt-cal`. The container accepts spans with the classes `rvt-cal__month` and `rvt-cal__day`. Both spans are optional, making it possible to use just the month, just the day, or both. For context, we prefer both month and day.
+
+```
+<div class="rvt-cal">
+  <span class="rvt-cal__month">
+  </span>
+  <span class="rvt-cal__day">
+  </span>
```
zttodd: Should we include values in this example (like Feb and 9), so that developers know what to include? I think this would be helpful to demonstrate that you use sentence case for the month and the case that's displayed on the page will be controlled by CSS.

src/sass/calendar/_base.scss
```diff
@@ -0,0 +1,34 @@
...
+    font-size: $ts-14;
+    text-transform: uppercase;
+    padding: 0 $spacing-sm;
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  &__day {
+    font-weight: $font-weight-regular;
+    font-size: $ts-16 * 1.5;
```
levimcg: I think you could get away with using a standard value from the type scale here instead of introducing a magic number. There are definitely times when that needs to happen, but in this case the visual difference between the result of this calculation (24px/1.5rem) and the standard `$ts-23` value (23px/1.4375rem) is so close that the trade-off might not be worth it.

src/sass/rivet.scss
```diff
@@ -13,6 +13,7 @@
 @use 'breadcrumb';
 @use 'buttons';
 @use 'buttons-segmented';
+@use 'calendar';
```
levimcg: We might think about calling this something like `calendar-tile`. My only concern is that it might be confused with an actual functional calendar/datepicker component.

PR #357, opened 2020-03-12
Added documentation for modal

illusivesunrae:
So this is definitely still a little rough. I would love feedback on the best syntax for listing the public methods now that we are using class instantiation. 

I did notice in the Aria authoring practices, it mentions that the modal container should have `aria-modal` set to `true`. This was introduced to replace `aria-hidden` for modals, but we are also not using that (we use `hidden` instead). We can decide if this is something we should create an issue for.

### DIFF
--- a/src/components/modal/README.md
+++ b/src/components/modal/README.md
@@ -4,7 +4,7 @@ A modal is a small window that is displayed on top of the main application/site 
 
 ## When to use
 
-Modals are very good for allowing users to handle small tasks related to an overarching workflow or goal. This could include:
+Modals are useful for allowing users to handle small tasks related to an overarching workflow or goal. This could include:
 
 - Updating personal information as part of a larger process
 - Determining information needed in their main workflow e.g. calculating a number, looking up a code
@@ -53,7 +53,7 @@ The modal component has two options that can be passed in on initialization - op
 
 ### openOnInit
 
-This option is intended to allow your modal to open on page load. Pass through the `openOnInit` option set to `true` when you instantiate the modal to enable this.
+This option is intended to allow your modal to open on page load. Pass in the `openOnInit` option set to `true` when you instantiate the modal to enable this.
 
 ```
 document.newModal = new Rivet.Modal(modalElement, { openOnInit: true });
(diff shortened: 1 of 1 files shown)


src/components/modal/README.md
```diff
@@ -1 +1,112 @@
-# Modal
\ No newline at end of file
+# Modal
+
+A modal is a small window that is displayed on top of the main application/site window. The main page is still visible, but the background is darkened to direct focus to the content of the modal window.
+
+## When to use
+
+Modals are very good for allowing users to handle small tasks related to an overarching workflow or goal. This could include:
```
zttodd: Should we change 'very good' to 'useful', for more neutral language?

src/components/modal/README.md
```diff
@@ -1 +1,112 @@
...
+
+- Updating personal information as part of a larger process
+- Determining information needed in their main workflow e.g. calculating a number, looking up a code
+
+## When to use something else
+
+Please consider using the alert component if you are in need of a way to display an error message. The modal component should not be used solely for this purpose.
+
+If you need to display large amounts of information for multiple tasks consider using a dedicated page instead.
+
```
zttodd: Is this a good opportunity to include a note discouraging developers from using the modal for email newsletter sign-up overlays?

src/components/modal/README.md
```diff
@@ -1 +1,112 @@
...
+document.newModal = new Rivet.Modal(modalElement);
+```
+
+## Options
+
+The modal component has two options that can be passed in on initialization - openOnInit and dialog.
+
+### openOnInit
+
+This option is intended to allow your modal to open on page load. Pass through the `openOnInit` option set to `true` when you instantiate the modal to enable this.
```
zttodd: We might not need the word "through" here. Could it say something like:

```suggestion
This option is intended to allow your modal to open on page load. Pass in the `openOnInit` option set to `true` when you instantiate the modal to enable this.
```

src/components/modal/README.md
```diff
@@ -1 +1,112 @@
...
+
+```
+document.newModal = new Rivet.Modal(modalElement, { openOnInit: true });
+```
+
+### dialog
+
+A modal dialog is similar to a regular modal except that it requires the user to interact with it before continuing any interaction with the main application. The user must select from the available actions in the dialog—they cannot disregard and simply close the window.
+
+Pass through the `dialog` option set to `true` when you instantiate the modal to enable this.
```
zttodd: Same comment as above ☝️

PR #358, opened 2020-03-12
Added documentation for the file input component

zttodd:
This component adds the following sections to the `file input` component docs:
- When to use
- Adding the markup
- Initialization
- Accessibility requirement

It would be helpful to have some feedback on how to improve the Initialization section. I had a difficult time thinking of a standard way to recommend initializing, and explaining it a way that could be applied to each developer's unique scenario, and also not under-explaining it for people who don't have as much JavaScript experience.

I also wasn't sure of an instance where you wouldn't use a file input, since it seems to have one clear function, but I'm sure I'm missing some other instances where it might be misused.

### DIFF
--- a/src/components/file-input/README.md
+++ b/src/components/file-input/README.md
@@ -23,8 +23,6 @@ The file input component is made up of a wrapping `div` with the class `rvt-file
 
 The elements `div[data-upload-wrapper]`, `input[data-upload-input]`, `input[id]`, and `div[data-upload-preview]` should all have the same unique `id`.
 
-The attribute `input[aria-describedby]` should match `div[id]` for assistive technology to properly read the file description.
-
 ```
 <div class="rvt-file" data-upload-wrapper="fileInputId">
     <input type="file" data-upload-input="fileInputId" id="fileInputId" aria-describedby="file-description">
@@ -66,7 +64,7 @@ If you use the appropriate data attribute/id combination in your markup, the fil
 
 ## Accessibility Requirements
 
-The Rivet file input is built to follow the WAI-ARIA authoring standards. It is marked up with the appropriate ARIA attributes and uses the JavaScript included in rivet.js to implement the keyboard navigation and focus management required to meet the ARIA Authoring Practices standards. If you need to create the file input functionality in another framework/library like React, Angular, etc., please ensure that it meets the following accessibility requirements.
+Follow these guidelines to ensure that the file input component meets accessibility requirements.
 
 ### Focus
 - The file input button should have a visible :focus state
--- a/src/components/file-input/README.md
+++ b/src/components/file-input/README.md
@@ -7,6 +7,8 @@ Use the file input component when the user is required to upload a file from the
 
 ## Adding the Markup
 
+The file input component is made up of a wrapper which has the following contents: `input`, `label`, and `div`.
+
 ### Wrapper
 
 The file input component is made up of a wrapping `div` with the class `rvt-file` and the data attribute `data-upload-wrapper`.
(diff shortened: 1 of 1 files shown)


src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
-# File input
\ No newline at end of file
+# File input
+
```
illusivesunrae: Probably good to add a general description of the file input component here.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+## When to use
+Use the file input component when the user is required to upload a file from their local computer.
+
+## Adding the Markup
```
illusivesunrae: I think this section has become a little too divided. More info below for details.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+
+### Wrapper
+
+The file input component is made up of a wrapping `div` with the class `rvt-file` and the data attribute `data-upload-wrapper`.
```
illusivesunrae: I think this could be broken down better. 

> The file input component is made up of a wrapping `div` with the class `rvt-file` and the data attribute `data-upload-wrapper`.

## Adding the markup

The file input component is made up of a wrapper and (list all the other required elements).

### Wrapper

The wrapper for the file input is a `div` with the class `rvt-file` and the data-attribute `data-upload-wrapper`.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+
+**`div`**:
+
+```
+<div class="rvt-file__preview" data-upload-preview="fileInputId" id="file-description">
+    No file selected
+</div>
+```
+
+The elements `div[data-upload-wrapper]`, `input[data-upload-input]`, `input[id]`, and `div[data-upload-preview]` should all have the same unique `id`.
```
illusivesunrae: It feels like this could be used to tie the elements section together more by introducing it near the top.

### Elements

The data elements in the elements below should all have the same unique `id`. This applies to `data-upload-wrapper`, `data-upload-input`, `id`, and `data-upload-preview`.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+
+The attribute `input[aria-describedby]` should match `div[id]` for assistive technology to properly read the file description.
```
illusivesunrae: If this is in the A11y section, you can probably remove it from here, otherwise, just move it to that section.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+If you use the appropriate data attribute/id combination in your markup, the file input will work without the need for any additional JavaScript. But if you need to control the file input programmatically, there are two methods from the Rivet file input's API you can use:
+
+| Method           | Description                                                                            |
+| ---------------- | -------------------------------------------------------------------------------------- |
+| .init()  | Adds the built-in event listeners to the file input |
+| .destroy() | Removes all built-in event listeners from the file input |
+
+## Accessibility Requirements
+
+The Rivet file input is built to follow the WAI-ARIA authoring standards. It is marked up with the appropriate ARIA attributes and uses the JavaScript included in rivet.js to implement the keyboard navigation and focus management required to meet the ARIA Authoring Practices standards. If you need to create the file input functionality in another framework/library like React, Angular, etc., please ensure that it meets the following accessibility requirements.
```
illusivesunrae: Make sure to add the link to the ARIA Authoring Practices standards webpage: https://w3c.github.io/aria-practices/

levimcg: I'm not sure this whole part about ARIA authoring practices is needed here. The file input is not any kind of officially documented widget that aria has recommended roles and states for. For example, if you look at the left nav on the ARIA site, there's no link to "file input" or equivalent. Other widgets like Accordion, Modal, Tabs, etc. are more "standar" widgets that ARIA has specs for. Eg. you can [link directly to the recommendations for Accordions](https://w3c.github.io/aria-practices/#accordion)

Our implementation of the file input is mostly custom styling with JavaScript that helps make it more accessible, but the ARIA doesn't necessarily make any recommendations here.

It's still important to have the accessibility requirements section here to call out the focus behavior, keyboard interactions, and any labeling, roles, states, etc. like what you have here for `aria-describedby`.

src/components/file-input/README.md
```diff
@@ -1 +1,99 @@
...
+    - `const newFileInput = new Rivet.FileInput(selectedFileInput)`
+
+## "Public" methods
+
+If you use the appropriate data attribute/id combination in your markup, the file input will work without the need for any additional JavaScript. But if you need to control the file input programmatically, there are two methods from the Rivet file input's API you can use:
+
+| Method           | Description                                                                            |
+| ---------------- | -------------------------------------------------------------------------------------- |
+| .init()  | Adds the built-in event listeners to the file input |
+| .destroy() | Removes all built-in event listeners from the file input |
```
levimcg: Nice, I'm not sure these were documented in Modal component PR (#357). We'll need to make sure `init` and `destroy` are documented for all components.

src/components/file-input/README.md
zttodd: Ah, I understand. I misinterpreted that portion from another component's documentation on rivet.iu.edu as an explanation that we've used `aria` attribute(s) to improve accessibility. This is great to know – I'll adjust it here and keep it in mind for future documentation 👍

levimcg: Yeah, i _think_ the file input in Rivet is the only JS component that is outlier. I believe every other Rivet JS component maps to a specific design pattern/widget in the aria authoring practices docs. 👍

zttodd:
@levimcg @illusivesunrae after making adjustments based on your feedback, this is ready for another review.

src/components/file-input/README.md
```diff
@@ -1 +1,76 @@
...
+Follow these guidelines to ensure that the file input component meets accessibility requirements.
+
+### Focus
+- The file input button should have a visible :focus state
+
+### Labeling
+- The `input` element should have the attribute `aria-describedby`, which should match the `id` on `div.rvt-file__preview`
+
+### Keyboard navigation
+- Enter or Space = Activate file input dialog (when focused)
```
illusivesunrae:
```suggestion
- `enter` or `space` = Activate file input dialog (when focused)
```

src/components/file-input/README.md
```diff
@@ -1 +1,76 @@
...
+| ---------------- | -------------------------------------------------------------------------------------- |
+| .init()  | Adds the built-in event listeners to the file input |
+| .destroy() | Removes all built-in event listeners from the file input |
+
+## Accessibility requirements
+
+Follow these guidelines to ensure that the file input component meets accessibility requirements.
+
+### Focus
+- The file input button should have a visible :focus state
```
illusivesunrae:
```suggestion
- The file input button should have a visible `:focus` state
```

PR #359, opened 2020-03-13
Added documentation for tabs

illusivesunrae:


src/components/tabs/README.md
```diff
@@ -1 +1,141 @@
...
+
+✅ Activating a tab should set its associated tab panel’s `tabindex` attribute to `-1`. All in active tabs should have their `tabindex` attribute set to `0` so that after the used activates a tab, pressing `tab` will move focus to the active tab’s tab panel.
+
+### Labeling
+
+✅ Tabs should be wrapped in an element with a role of tablist.
+
+✅ Each tab (button) should have a role of tab.
+
+✅ Each tab panel should have a role of tabpanel.
```
zttodd:
```suggestion
✅ Tabs should be wrapped in an element with a `role` of tablist.

✅ Each tab (button) should have a `role` of tab.

✅ Each tab panel should have a `role` of tabpanel.
```

src/components/tabs/README.md
```diff
@@ -1 +1,141 @@
...
+  <div class="rvt-tabs__panel" tabindex="0" role="tabpanel" id="tab-3" aria-labelledby="t-three" data-tab-panel="tab-3" hidden data-tab-init>
+    <span class="rvt-ts-23 rvt-text-bold">A grid inside a tab panel</span>
+    <p>Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+  </div>
+</div>
+```
+
+## Initialization
+
+In order to initialize the modal as an ES6 module, you will need to create a new instance of it, and pass it the element you wish to instantiate as a set of tabs.
```
levimcg: Same as the modal docs here. You don't need to call out using it specifically as an ES module if you're not documenting how to import it first. The instructions for instantiation and code example here is the same for whether you are using the ES module or using the iife versionn in a script tag. See comments in #357 for more info.

Also just a quick note on style. We can refer to modules as plain _ES modules_ or _ECMAScript_ modules (a little wordy tho). ES6 was the particular version of the ECMAScript specification where modules were standardized. The current version is ES8, for example. Modules will be in every version moving forward, so there's no need to specify _ES6_ specifically. 👍

src/components/tabs/README.md
```diff
@@ -1 +1,141 @@
...
+
+```
+const tabsElement = document.querySelector('[data-tabset="tabset-1"]');
+document.newTabs = new Rivet.Tabs(tabsElement);
```
levimcg: Same comments as #357 here. No need to show creating a property on the `document` object here. Same will apply to all code examples.

```suggestion
const newTabs = new Rivet.Tabs(tabsElement);
```

src/components/tabs/README.md
```diff
@@ -1 +1,141 @@
...
+```
+
+## "Public" methods
```
levimcg: We can remove the quotes here on _Public_ and in all the other docs. I think I may have written it that way somewhere as because I was inferring that technically JS classes don't implement a true concept of `public` vs. `private` methods and properties like other class-based languages do, [at least not yet anyway](https://github.com/tc39/proposal-private-fields).

Technically _all_ fields and methods are _public_ in JS classes, but we using naming conventions like the leading underscore (`_somePrivateMethodOrField`) to help communicate to other developers that a method is for internal use only. Eg. is should not be called from outside.

It might be more appropriate to standardize calling this section of Rivet components that use JS something like: `JavaScript API`.
```suggestion
## JavaScript API
```

src/components/tabs/README.md
```diff
@@ -1 +1,141 @@
...
+
+## Custom events
+
+| Event          | Description                                                                                                                                                                                                                                                                                                                        |
+| -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `tabActivated` | Emitted when an individual tab panel is activated (using the `Tabs.activateTab()` method, or via a click on a button with the `data-tab` attribute). The value of the `data-tab` attribute is also passed along (if it exists) via the custom event’s detail property and is available to use in your scripts as event.detail.id() |
```
levimcg: I believe in Rivet 1.x.x javascript the `event.detail.id` was a function that returned the id, which is why the old docs where written `event.detail.id()` (calling the function). With the new JS, we have simplified it so it's a plain property on the `event.detail` object.
```suggestion
| `tabActivated` | Emitted when an individual tab panel is activated (using the `Tabs.activateTab()` method, or via a click on a button with the `data-tab` attribute). The value of the `data-tab` attribute is also passed along (if it exists) via the custom event’s detail property and is available to use in your scripts as event.detail.id |
```

PR #360, opened 2020-03-13
Added documentation for dropdown

zttodd:
This PR adds a first draft of the docs for the `dropdown` component. One item to note is in the **labeling** section, there is an _optional_ `aria-controls` that can be added. We haven't included that within our Fractal/Nunjucks template, so I wanted to make sure that this was something we still wanted to include within the docs, and if so whether we should add it into the Nunjucks template.

I borrowed the table layout and language from #357 for the **"public" methods** and **custom events** sections.

### DIFF
--- a/src/components/dropdown/README.md
+++ b/src/components/dropdown/README.md
@@ -1,4 +1,5 @@
 # Dropdown
+
 Use the dropdown component to create a list of menu options that can be toggled with a button element.
 
 ## When to use
@@ -8,3 +9,11 @@ Use a dropdown menu when you need to give users a list of actions or links to ch
 ## When to use something else
 
 Although similar to a native HTML <select> element, the dropdown component should not be used as a replacement inside forms. Use the select element instead when you need to give users a list of mutually exclusive choices while filling out a form.
+
+## Initialization
+
+## "Public" methods
+
+## Custom events
+
+## Accessibility requirements
--- a/src/components/dropdown/README.md
+++ b/src/components/dropdown/README.md
@@ -1,2 +1,10 @@
 # Dropdown
 Use the dropdown component to create a list of menu options that can be toggled with a button element.
+
+## When to use
+
+Use a dropdown menu when you need to give users a list of actions or links to choose from.
+
+## When to use something else
+
+Although similar to a native HTML <select> element, the dropdown component should not be used as a replacement inside forms. Use the select element instead when you need to give users a list of mutually exclusive choices while filling out a form.
--- a/src/components/dropdown/README.md
+++ b/src/components/dropdown/README.md
@@ -112,7 +112,7 @@ If you use the appropriate data attribute/id combination in your markup, the dro
 
 ## Accessibility requirements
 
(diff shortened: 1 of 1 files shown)


src/components/dropdown/README.md
```diff
@@ -1 +1,138 @@
...
+## Initialization
+
+In order to initialize the dropdown as an ES6 module, you will need to create a new instance of it, and pass it the element you wish to instantiate as a dropdown.
+
+```
+const dropdownElement = document.querySelector('[data-dropdown="your-data-dropdown-value"]');
+document.newDropdown = new Rivet.Dropdown(dropdownElement);
+```
+
+## "Public" methods
```
levimcg: See my comments on similar section in #359 about use of "Public" in quotes.
```suggestion
## JavaScript API
```

src/components/dropdown/README.md
```diff
@@ -1 +1,138 @@
...
+| .destroy() | Removes all built-in event listeners from the dropdown |
+
+## Custom events
+
+| Event        | Description                                                                                                                                                                                                                                                          |
+| ------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `dropdownOpen`  | Emitted when the dropdown is opened (using the Dropdown.open() method). The value of the dropdown id attribute is also passed along via the custom event’s detail property and is available to use in your scripts as event.detail.id()  |
+| `dropdownClose` | Emitted when the dropdown is closed (using the Dropdown.close() method). The value of the dropdown id attribute is also passed along via the custom event’s detail property and is available to use in your scripts as event.detail.id() |
```
levimcg: See same comment in #359 about the `event.detail` function vs. regular object property. Should be able to document as `event.detail.id` without parens.

```suggestion
| `dropdownClose` | Emitted when the dropdown is closed (using the Dropdown.close() method). The value of the dropdown id attribute is also passed along via the custom event’s detail property and is available to use in your scripts as event.detail.id |
```

src/components/dropdown/README.md
```diff
@@ -1 +1,138 @@
...
+
+## Accessibility requirements
+
+The Rivet dropdown is built to follow the WAI-ARIA authoring standards. It is marked up with the appropriate ARIA attributes and uses the JavaScript included in `rivet.js` to implement the keyboard navigation and focus management required to meet the [ARIA Authoring Practices standards](http://w3c.github.io/aria-practices/). If you need to create the dropdown functionality in another framework/library like React, Angular, etc., please ensure that it meets the following accessibility requirements.
```
levimcg: This is a small detail, but since we now ship multiple versions of Rivet JS, you might need to update some language here. For example, we don't ship a JS file called `rivet.js` anymore since there is an ESM and IIFE version.

src/components/dropdown/README.md
```diff
@@ -1 +1,137 @@
...
+</div>
+```
+
+## Initialization
+
+In order to initialize the dropdown as an ES6 module, you will need to create a new instance of it, and pass it the element you wish to instantiate as a dropdown.
+
+```
+const dropdownElement = document.querySelector('[data-dropdown="your-data-dropdown-value"]');
+document.newDropdown = new Rivet.Dropdown(dropdownElement);
```
illusivesunrae:
```suggestion
const newDropdown = new Rivet.Dropdown(dropdownElement);
```

PR #365, opened 2020-03-18
Migrated timeline documentation

illusivesunrae:


src/components/timeline/README.md
```diff
@@ -1,12 +1,115 @@
...
+    </div>
+  </div>
+</div>
+```
+
+### External date labels
+
+The `rvt-timeline__date--label` class can be added to labels to make them appear on the opposite side of the timeline from the main pieces of content. This only works for certain screen sizes, but is compatible with all three alignment options.
+
+The external labels are set to a width of 10rem in order to create a uniform appearance. To adjust this, use Rivet width utility classes on the `span` with the `rvt-timeline__date--label` class.
```
zttodd:
```suggestion
The external labels are set to a width of `10rem` in order to create a uniform appearance. To adjust this, use Rivet width utility classes on the `span` with the `rvt-timeline__date--label` class.
```

src/components/timeline/README.md
```diff
@@ -1,12 +1,115 @@
...
 
 ## When to use
-- Your website or app needs to display a series of time-related events/activities that require further explanation
 
-## When to consider something else
-- When communicating where users are in a process, use the Step Indicator
-- For normal lists, use ordered lists or unordered lists
+The timeline component should be used to structure chronologically relevant information in a visually meaningful way. This can include:
+
+- Displaying a series of time-related events/activities that require further explanation
```
zttodd: Since this list only has one item, would it be better to combine it all into one or two paragraphs?

PR #369, opened 2020-03-20
Migrated documentation for link component

zttodd:
This PR migrates the `link` docs, adds a markup section, and reworks the "Microcopy & accessibility notes" into "accessibility requirements" and "implementation notes".

### DIFF
--- a/src/components/link/README.md
+++ b/src/components/link/README.md
@@ -7,12 +7,16 @@ Inline text links are used to navigate between documents (pages).
 - Navigating to a new page or view in your application
 - Navigating to different web page, e.g. external documentation
 
-## When to consider something else
+## When to use something else
 
 Use the button component for:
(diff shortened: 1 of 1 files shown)


src/components/link/README.md
```diff
@@ -1 +1,44 @@
...
+## When to use something else
+
+Use the button component for:
+- Opening or closing a modal or dialog
+- Triggering a dropdown menu
+- Submitting data to the server
+
+## Adding the markup
+
+To mark up a link in Rivet, add the `a` HTML element.
```
illusivesunrae: Should we be encouraging people to use just `a` tags, or `a` tags with the class `rvt-link`? I know we allow the use of either.

zttodd: Good catch! I've added a note and example using the `rvt-link` class. Let me know what you think!

PR #371, opened 2020-03-20
Migrated documentation for pagination component

zttodd:
This PR migrates the docs for the paginations component, adds a markup section, and reorganizes and renames the sections.

### DIFF
--- a/src/components/pagination/README.md
+++ b/src/components/pagination/README.md
@@ -1,1 +1,4 @@
 # Pagination
+
+The pagination component is used to break up large sets of data across multiple pages.
+
--- a/src/components/pagination/README.md
+++ b/src/components/pagination/README.md
@@ -2,6 +2,14 @@
 
 The pagination component is used to break up large sets of data across multiple pages.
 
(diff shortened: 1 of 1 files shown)


src/components/pagination/README.md
```diff
@@ -1 +1,228 @@
...
+    </li>
+    <li class="rvt-pagination__item">
+      <a href="#" aria-label="Page 20, last page">20</a>
+    </li>
+    <li class="rvt-pagination__item">
+      <a href="#" aria-label="Next set of pages">Next</a>
+    </li>
+  </ul>
+</nav>
+```
```
illusivesunrae: I think this is fine for now, but we may want to consider standardizing how we structure the `## Adding the markup` sections. I feel like we have some where we just list the default under the section title, and then add a subsection for Additional options, and sometimes we don't.

zttodd: Good call. I've made a note to re-evaluate these docs for consistent structure once we've worked through them all.

PR #374, opened 2020-03-23
Migrated documentation for breadcrumbs

zttodd:


src/components/breadcrumb/README.md
```diff
@@ -1 +1,50 @@
...
+</nav>
+```
+
+## When to use
+
+Use breadcrumbs when your application has multiple levels in the navigation hierarchy. Breadcrumbs can help users navigate quickly between these levels.
+
+## When to use something else
+
+Avoid using the breadcrumb nav when your application only has one level of navigation.
```
illusivesunrae: For consistency, move these above `## Adding the markup`.

PR #380, opened 2020-03-24
Migrated docs for the checkbox component

zttodd:


src/components/checkbox/README.md
```diff
@@ -1 +1,144 @@
...
+                <label for="checkbox-5">Option four checked and disabled</label>
+            </li>
+        </ul>
+    </fieldset>
+</form>
+```
+
+### Indeterminate State
+
+It can be used to indicate a partially checked state. A typical use case would be nested checkboxes/selections like the following example. [This article](#) explains the indeterminate state in greater detail. **You will need to initialize the indeterminate JavaScript in order to use this functionality.** [UPDATE LATER WITH INSTRUCTIONS FOR HOW TO DO THIS]
```
illusivesunrae:
```suggestion
Indeterminate state is a built-in checkbox property that allows you to indicate a partially checked state. A typical use case would be nested checkboxes/selections like the following example. [This article](#) explains the indeterminate state in greater detail. **You will need to initialize the indeterminate JavaScript in order to use this functionality.**

To initialize the JavaScript for this, you will need to retrieve the element and set the indeterminate state to `true`.

```
// Initializing indeterminate state

```
<script>
const checkbox = document.querySelector('#checkbox-indeterminate');
    checkbox.indeterminate = true;
</script>
```

src/components/checkbox/README.md
```diff
@@ -1 +1,144 @@
...
+        <li>
+            <input type="checkbox" name="checkbox-demo" id="checkbox-7">
+            <label for="checkbox-7">Parent option two</label>
+        </li>
+        <li>
+            <input type="checkbox" name="checkbox-demo" id="checkbox-8">
+            <label for="checkbox-8">Parent option three</label>
+        </li>
+    </ul>
+</fieldset>
```
illusivesunrae: Wrap in backticks.

src/components/checkbox/README.md
illusivesunrae: The code suggestion logic has a bit of an issue with items wrapped in backticks within the suggestion, so this is a little messed up. Everything below the visual suggestion should be wrapped in backticks.

PR #398, opened 2020-09-23
Added new header system

levimcg:
This PR adds a standard version of the new header system in use across various web properties created over the past few months. [Refer to this RFC](https://v2.rivet.iu.edu/rfcs/header-system/) for more detailed info and documentation.

This is still a work in progress. I'll work to identify some specifics that need refactored but here are some outstanding issues that need solved:

- See the `/sass/header-system/_shame.scss` file. This `-global-toggle` elements gets used in multiple pieces/modules of the header, so it's more of a stand-alone component. There are lots of contextual variants based on what function it controls e.g. open/close, search toggle, etc. This could be refactored to be more generic, I think
- Initialization/setup logic for JavaScript in the header. there are some specifics to how the different types of toggles/menus are being initialized in the header. See `header-system.njk` for more details. Should we create a new `Header.js` javascript component that handles this and any other one-off stuff we need to think about in the header? Just a thought.
- there are still some minor fit-and-finish details to address. Things like inconsistent space, alignment, etc.
- Still need to write some basic docs to at least outline how all the different pieces work together.

@ohnoitsaustin feel free to have a look as we discussed. I'm interested to get your opinion from a developer experience perspective. This thing is pretty complicated.


scottanthonymurray:
@levimcg I've pushed some more commits to this branch that extracts dropdown-specific code into its own `Dropdown` subclass that extends a new parent class `Disclosure`.

When you get a few spare moments, can you pull this branch down and confirm that the header system still works correctly? It seems to work correctly in Fractal but I want to make sure I haven't missed anything.

levimcg:
@scottanthonymurray Yep, will do. I should have some time this afternoon.

levimcg:
@scottanthonymurray This is looking really good. creating the base disclosure class and extracting the dropdown functionality into a new subclass definitely cleans it up a lot and makes it soooo much more readable, imo. 💯 

I think there are a couple of outstanding things we should discuss:

1. I'd like to sort of define what the move to web components vs. the traditional DOM manipulations stuff we have going on right now looks like. That way we can be on the same page for how we want to approach that work. I think we should probably get aligned no what the transition [you outline in comments here](https://github.com/indiana-university/rivet-source/pull/398/files#diff-45002d467a676b1dd272afafac370f6e947775f9882c6afdc0092cec8d8da5e3R18) looks like.
1. In the [comment linked above](https://github.com/indiana-university/rivet-source/pull/398/files#diff-45002d467a676b1dd272afafac370f6e947775f9882c6afdc0092cec8d8da5e3R18) you mention "backwards compatibility for v1.x.x. ..." I'd also like to get aligned on this idea. We're working on the 2.0.0-alpha release track here so we don't need to worry about so much about backwards compatibility for 1.x releases. We might need to be thinking about what was released in the [2.0.0-alpha.3](https://github.com/indiana-university/rivet-source/releases/tag/v2.0.0-alpha.3) (last v2 alpha release) in terms of compatibility, but even there `alpha` infers a "use at your own risk" situation in the semantic versioning world. Meaning we can make breaking changes between alpha releases, especially since everything is undocumented right now.  
1. More generally I'd like to talk in a little deeper detail about [the initialization logic here](https://github.com/indiana-university/rivet-source/pull/398/files#diff-4641b48a0c943f05fd9c02c1669d9bf4ab981ca82466cc5372ab67d3ed6ccc65R7). If we're moving to web components the initialization logic point becomes moot, but until then does it make sense to put the writing of the setup work (querying the DOM and initializing dropdowns vs. disclosures based on data attributes) on the developer? I don't necessarily expect us to answer those questions in this PR, but if we are going to release, for example, a `2.0.0-alpha.4` we'll probably need to get these things on our collective radar.

Maybe we can start to dig into this in our weekly touch base tomorrow?

levimcg:
@scottanthonymurray I'm done with the updates I wanted to make to the PR. I think it's time to merge and then we can break up any remaining work/refinement into smaller PRs. I've made some notes in the code and created an Asana tasks to keep track of some updates/decisions that eventually need made, but I'm happy with this a first pass at an alpha version of the header system. If you don't have any objections, I'll go ahead and merge.

PR #436, opened 2021-02-15
Updated button focus outline color to match border/fill color

### LINKED ISSUE #388: Contextual focus styling
This may be more of a personal style preference thing, but is there any reason why the focus outline styling on buttons is always the primary color, and not the actual context state color? For example, buttons styled with the 'danger' classes create a red button, but with blue focus styling:
buttonfocus

scottanthonymurray:
Fixes #388.


levimcg (approved):
Looks good to me. 🚀

PR #437, opened 2021-02-15
Added toggle variable check to dropdown component JS

### LINKED ISSUE #319: Dropdown error when loaded and then later deleted from page
## Describe the bug
cc: @illusivesunrae @zttodd This bug occurs in Rivet `1.7.2`.

This bug was reported on Slack. We were able to troubleshoot and come up with a temporary fix, this is something we should look into patching. From Slack conversation:

> My team has an issue with Rivet drop downs. The drop down element exists on the page, is opened, is closed, then the drop down element is deleted. The following clicks on the page result in this console error: "rivet.min.js:8 Uncaught TypeError: Cannot read property 'hasAttribute' of null". That happens because of the following code in the Rivet JavaScript closeMenu function:

```js
// Return if disabled dropdown is being closed programmatically
if (toggle.hasAttribute('disabled')) {
  return;
}
```

And a proposed solution:

> The Rivet drop down _handleKeydown function has the following code which would also solve the issue (it would reset activeMenu so there are no errors after the drop down menu is deleted):
```js
case KEYS.escape:
// If there's an open menu, close it.
if (activeMenu) {
  closeMenu(activeMenu);
}
if (activeToggle && activeToggle !== null) {
  activeToggle.focus();
}
/**
 * Resets the state variables so as not to interfere with other
 * Escape key handlers/interactions
 */
activeMenu = null;
```

## Expected behavior
Keydown handler should check for the existence of the dropdown toggle and/or menu and reset the `activeMenu` variable to null if either aren't on the page anymore.

Solution 1:
Instead of calling window.Dropdown.init() once for all dropdowns, call window.Dropdown.init(context) and window.Dropdown.destroy(context) for every drop down we use. This solution would allow us to still use Rivet but is tedious and would break easily.
Solution 2:
Write our own custom drop down instead of using the Rivet drop down for situations Rivet has errors.
Solution 3:
The Rivet drop down _handleKeydown function has the following code which would also solve the issue (it would reset activeMenu so there are no errors after the drop down menu is deleted):

## Steps to reproduce
Add steps to reproduce the bug.

1. Add a dropdown menu to the DOM and initialize
1. Open and or close it
1. Remove it from the DOM (without re-initializing)
1. Click anywhere on the page, or use the escape key.

scottanthonymurray:
Fixes #319.

### DIFF
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -58,12 +58,12 @@ var Dropdown = (function() {
     var toggle = document.querySelector(toggleSelector);
 
     // Return if disabled dropdown is being opened programmatically
-    if (toggle && toggle !== null && toggle.hasAttribute('disabled')) {
+    if (toggle && toggle.hasAttribute('disabled')) {
       return;
     }
 
     // If the menu was opened by clicking an associated toggle
-    if (toggle && toggle !== null) {
+    if (toggle) {
       toggle.setAttribute('aria-expanded', 'true');
 
       activeToggle = toggle;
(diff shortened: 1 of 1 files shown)


levimcg (approved):
Nice, that fixed the bug for me.

I think this would be a good opportunity to add an addition integration test for this scenario in the `cypres/integration/dropdown_spec.js` tests. Cypress is pretty straightforward to use, but I'd be happy to sit down together for 1/2 hour this week and run through adding this test if you want?

I'd like to get that added before the next release, but let's maybe do that as a separate PR. We can talk about it 1:1 later this week.

scottanthonymurray:
@levimcg Sure thing, that sounds great!

PR #445, opened 2021-02-24
Added GitHub workflow for v1.x.x releases

scottanthonymurray:
@levimcg I _think_ this should be everything we need set up ahead of tomorrow's release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "start": "gulp",
     "headless": "gulp headless",
     "build": "npm run clean && gulp release && gulp fractalBuild && zip -r rivet.zip css js sass index.html",
-    "asset:upload": "npm run clean && gulp release && zip -r rivet-gh-release.zip css js sass tokens index.html",
+    "asset:upload": "npm run clean && gulp release && zip -r rivet-gh-release.zip css js sass index.html",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "start": "gulp",
     "headless": "gulp headless",
     "build": "npm run clean && gulp release && gulp fractalBuild && zip -r rivet.zip css js sass index.html",
+    "asset:upload": "npm run clean && gulp release && zip -r rivet-gh-release.zip css js sass tokens index.html",
```
levimcg, line 19: We haven't implemented the design tokens stuff on the 1.x.x develop and master branches yes (only on 2.0.0-dev/master), so you'll need to remove the `tokens` directory from the list of directories and files we're zipping up here toward the end of this line.

A bit of a tangent , but it may make sense to call this task something a something a little more descriptive. I keep thinking that this task actually _uploads_ some asset when it actually is just building/compressing the files into an asset that will eventually be uploaded.

levimcg (changes requested):
One small change here, but otherwise looks great! 💯

package.json
scottanthonymurray: @levimcg Added 2 more commits to address these. 📋

levimcg (approved):
Looks good 💯

PR #449, opened 2021-03-10
Added footer system

scottanthonymurray:
This pull request adds the HTML and Sass for the footer system as seen on [iu.edu](https://www.iu.edu/index.html).


src/components/includes/includes--footer-related.njk
```diff
@@ -0,0 +1,84 @@
...
+        link: "https://jobs.iu.edu/"
+      },
+      {
+        title: "Non-discrimination Notice",
+        link: "/nondiscrimination/index.html"
+      }
+    ]
+  }
+] -%}
+<div aria-labelledby="resources-heading" class="rvt-c-footer-related" role="complementary">
```
levimcg, line 56: I think we can maybe do a better job of naming these elements of the footer with more clear/semantic names. For example "related" isn't super clear to me. Related to what?

definitely not something I want to hold this PR up for, but something I want to figure out before we release the next alpha. that's something I'm happy to take on. I'll create an internal task for it.

src/components/includes/includes--footer-social.njk
```diff
@@ -0,0 +1,65 @@
+<div aria-labelledby="social-media-heading" class="rvt-c-footer-social" role="complementary">
```
levimcg, line 1: This doesn't need to hold things up right on this PR for now either, but the `rvt-c-...` convention was only used to denote that this was a "custom" component/implementation and not a built in part of the rivet 2 alpha we were using.

All the `-c-` suffixes can be removed when we move into the real code base. 👍🏻  I think we can tackle this as a separate refactoring pass/PR though.

levimcg (approved):
I think this looks good. 💯 

I'm going to create a couple of new tasks to make sure we come back for a finer-grain refactoring where we can focus a bit more on naming things, etc. (see review comments)

PR #456, opened 2021-04-01
Rewrote component JS to use wicked-elements

scottanthonymurray:
This pull request rewrites the JavaScript-powered components to use the `wicked-elements` library.


src/components/tabs/tabs.njk
```diff
@@ -1,35 +1,35 @@
-<div class="rvt-tabs" data-tabset="tabset-1">
+<div class="rvt-tabs" data-rvt-tabs="tabset-1">
     <div class="rvt-tabs__tablist" role="tablist" aria-label="Rivet tabs">
         <button class="rvt-tabs__tab"
```
levimcg, line 3: As a separate refactoring pass, I'd like to take a look and see if we can do a bit more to focus on taking some of the set-up work off of the developers writing the markup for these components.

For example adding all the roles, states and attributes related to accessibility and interactivity. Could we do that setup when the element is connected to the DOM? That way the markup the developer needs to grab is much simpler and less error prone when setting up in their own templates.

These tab buttons/panels are a good example. we could probably just add/set up all the aria roles and states in our component code instead of the developer having to make sure they're there and matched up correctly, etc.

src/components/tabs/tabs.njk
```diff
@@ -1,35 +1,35 @@
...
                 role="tab"
                 aria-selected="false"
-                data-tab="tab-4"
+                data-rvt-tab="tab-4"
                 id="t-four">
             Tab four
         </button>
     </div>
-    <div class="rvt-tabs__panel" tabindex="0" role="tabpanel" id="tab-1" aria-labelledby="t-one" data-tab-panel="tab-1" hidden>
+    <div class="rvt-tabs__panel" tabindex="0" role="tabpanel" id="tab-1" aria-labelledby="t-one" data-rvt-tab-panel="tab-1" hidden>
```
levimcg, line 32: Similar story here. Following some progressive enhancement principles might be the way to go, where we, the component authors, add the hidden attributes in our set up code. that way if something goes wrong on the client or in someone's app, the content is still accessible even though the toggle/tabs controls might not be working.

src/js/components/alert.js
```diff
@@ -2,55 +2,57 @@
...
+        Component.bindMethodToDOMElement(this, 'dismiss', this.dismiss);
+        
+        this._handleClick = this._handleClick.bind(this);
+      },
+
+      connected() {
+        Component.dispatchComponentAddedEvent(this.element);
+
+        this.closeButton = this.element.querySelector(this.closeSelector);
+        this.closeButton.addEventListener('click', this._handleClick, false);
```
levimcg, line 30: We can take another pass at this, but click handlers like this should probably be moved to a native wicked-elements `onclick` method since it takes care of set up and tear down when the event listeners are added that way.

I think the only time we need to set up event listeners like this is when they are on something other that the component/element itself. So, say the `document` when we're listening for clicks outside the element itself.

src/js/components/sidenav.js
```diff
@@ -3,125 +3,121 @@
...
+
+      disconnected() {
+        Component.dispatchComponentRemovedEvent(this.element);
+      },
+
+      onClick(event) {
+        this._handleClick(event);
+      },
+
+      _handleClick(event) {
```
levimcg, line 79: Similar to some of the other event listener comments. Since this method only ever really gets used in set up and tear down, I think it makes sense to just move the body of the `_handleClick(event)` function into a wicked-element `onclick(event)` method so we get the setup and tear down for free.

I'll write this up a separate refactoring pass though.

src/js/components/sidenav.js
```diff
@@ -3,125 +3,121 @@
...
+
+        // bind methods
+        Component.bindMethodToDOMElement(this, 'open', this.open);
+        Component.bindMethodToDOMElement(this, 'close', this.close);
+
+        this._handleClick = this._handleClick.bind(this);
+      },
+
+      connected() {
+        Component.dispatchComponentAddedEvent(this.element);
```
levimcg, line 68: I think I get what's going on here, but I wonder if since we can't really make a call to the parent class method because of the context of `this`. Does it make sense to still make some of these methods members of the parent `Component` class if we're just calling them from the imported object instead of `super.dispatchComponentAddedEvent()`?

I wonder if methods like this make more sense as a utility function that gets imported and then used as needed. This isn't any kind of deal-breaker for me. Just spitballing.

levimcg (approved):
Left some comments here. I'll create separate more detailed to-dos to track some of the work I mentioned here.

Really nice work on converting these over 💯

PR #459, opened 2021-04-05
Added footer system documentation

scottanthonymurray:
This PR adds some basic documentation about the new footer system to Fractal.


src/components/footer-system/README.md
```diff
@@ -1,26 +1,254 @@
...
+
+Branding guidelines require each site to have an IU-branded footer. For more information on web branding guidelines, visit [brand.iu.edu](brand.iu.edu).
+
+## Usage
+
+A footer is made up of three parts: the **base footer**, the **resources module**, and the **social module**.
+
+> Only the base footer is required. The resources and social modules are optional.
+
+### Base footer (required)
```
levimcg, line 15: I stubbed out a "--light" variant for the footer base module which makes the background (of that module only) white and makes the text black. This may or may not be something we want to look at for all three modules, but for now I made the light variant of the base module as a direct replacement/alternative for the existing stripped-down/application-focused Rivet 1.x footer. See screenshot:

image

We may want to make a note about that here, just so it's on our radar to include in any alpha documentation we push out. You can add the `rvt-footer-base--light` modifier class to see how it works.

image

src/components/footer-system/README.md
```diff
@@ -1,26 +1,254 @@
...
\ No newline at end of file
+## Accessibility
+
+The footer is built to follow the WAI-ARIA authoring standards. It is marked up with the appropriate ARIA attributes, described below.
+
+### ARIA attributes
+
+|Attribute|Description|
+|-|-|
+|`role="complementary"`|Applied to the container `div` for the optional resources and social modules|
```
levimcg, line 252: You've written it further up in the docs and code examples, but it may be worth mentioning here in the description of the complementary role, that the elements that use this role must be siblings of the `<main>` element. Eg. NOT nested inside it, or inside the `<footer>` (base module) element. Just to reiterate that point home.

levimcg (approved):
@scottanthonymurray I left a couple of comments here. No need for another review once you add the two small notes I mentioned. Nice work! docs are clear and concise. 💯

src/components/footer-system/README.md
scottanthonymurray: Great! I'll add a note about this to the docs.

src/components/footer-system/README.md
scottanthonymurray: Can do!

PR #460, opened 2021-04-05
Removed unused visual regression testing packages and files

scottanthonymurray:
Uninstalled `pixelmatch` and `pngjs`.

Ran the test suite and confirmed no new issues cropped up as a result of uninstalling these libraries.


levimcg (approved):
Nice, thanks! 🪓

PR #461, opened 2021-04-05
Removed CircleCI

scottanthonymurray:
Removed CircleCI configuration and badge from repo, as we're no longer using it.


levimcg (approved):
💯

PR #462, opened 2021-04-06
Added header system documentation

scottanthonymurray:
This PR adds basic documentation for the new header system.


src/components/header-system/README.md
```diff
@@ -1,14 +1,384 @@
...
+
+```html
+<header class="rvt-header-wrapper">
+    
+    <!-- Hidden link for screen reader users to skip to main content -->
+    
+    <a class="rvt-header-wrapper__skip-link" href="#main-content">Skip to main content</a>
+    
+    <div class="rvt-header-global">
+        <div class="rvt-container-xl">
```
levimcg, line 27: I think we should have at least a small section/heading that covers the use of the generic `rvt-container-X` utility to constrain the max width of a site/app header.

In iu.edu we used the `rvt-container-xl` size, because of the number of navigation items, etc. The `-xl` container gave us the space we needed, but the point of using the container-x utility in the header is to allow developers/designers to set a max-width to their header to match the rest of the `rvt-container-x` elements they are using to constrain the max-width of their sites.

For example if I was building a web app that had a giant table that listed a bunch of record with lots of columns, I would probably want to use an `rvt-container-xl` to constrain the max width of the main content area (where the table lives) so as to give desktop users a lot space for the table to spread out on large monitors. On the other hand if I had a smaller site with some editorial content I might want to constrain the max width of my content to a `rvt-container-lg` element. This way I can just update the container element in the header of my site to match.

Not sure if that is making sense the way I'm describing it?

src/components/header-system/README.md
```diff
@@ -1,14 +1,384 @@
...
+
+### Implementing search
+
+By default, the user is sent to the URL `/search` when they submit a search query. The value of their search query is appended to the page URL as the query string parameter `q`.
+
+```
+GET /search?q={search query}
+```
+
+It is up to you to implement a method for handling search queries on your website or application &mdash; most sites use [Google Custom Search](https://developers.iu.edu/resources/google-custom-search/).
```
levimcg, line 252: This is just more of life hack 😃 , but you can use the key combo of `Alt` + `Shift` + `-` (minus key) to create an em dash character instead of having to type out the HTML entity by hand. As long as the font rendering the text has an em dash character (almost all fonts do), it will get rendered correctly. 🔥

src/components/header-system/README.md
```diff
@@ -1,14 +1,384 @@
...
+
+If you'd like search to function differently, you can:
+
+- Change the value of the `action` attribute on the search `form` element to redirect the user to a different page on submit 
+- Change the value of the `method` attribute on the search `form` element to `post` to submit the search query via `HTTP POST` instead
+- Change the value of the `name` attribute on the search `input` element to use a different name for the variable containing the search query
+
+To handle submitted search queries yourself without redirecting the user to a new page, use the following JavaScript:
```
levimcg, line 260: More of just a comment here, but I'm wondering if we need to really go into detail about search implementation specific to JavaScript?

I'm thinking specifically in the web apps. All we need to do is give them the markup, styles and the show/hide logic, and then the search implementation will very much depend on their situations. Some teams might just straight handle it all on the server. In front-end framework world I would imagine they'll handle it the way that fits with their framework.

In the iu web framework/websites world I would imagine the go to implementation will remain Google custom search.

I think it's fine to leave for now. Just wanted to get your thoughts.

src/components/header-system/README.md
```diff
@@ -1,14 +1,384 @@
...
+
+  const query = searchField.value;
+
+  // Handle query...
+});
+```
+
+## Secondary navigation
+
+You can add a second level of navigation links beneath the header. This option is useful for section-specific navigation or subsites within a larger website.
```
levimcg, line 277: I don't have a problem with this language, but I kind of remember purposefully removing all "subsite" language when we were writing the RFC for the header system last year. Let me see if I can find out why we made those changes. We don't need to let that hold up this PR though.

levimcg (approved):
@scottanthonymurray These docs are looking great to me. 💯  I left some comments here about a few things I'd like to figure out how to address and document as we move these docs into the actual docs site, but nothing that should stop this PR from moving forward. Great work! 🙌🏻

src/components/header-system/README.md
scottanthonymurray: Force of habit. Thanks for the tip!

src/components/header-system/README.md
scottanthonymurray: @levimcg Looking back at the "Implementing search" section with fresh eyes, I'm thinking of just getting rid of it. Presumably anyone building a site or app with Rivet understands basic HTML, can look at the markup for search, and understand how it's working. I can't imagine a developer using Rivet who isn't familiar with form HTML.

src/components/header-system/README.md
scottanthonymurray: @levimcg Oh interesting, I had no idea. I'll leave it in for now and we can revisit later.

src/components/header-system/README.md
scottanthonymurray: @levimcg It does, thanks!

PR #468, opened 2021-04-21
Fixes error when Modal.open or Modal.close is called on a modal that has been removed from the DOM

### LINKED ISSUE #451: Modal.close(id) sometimes throws an undefined exception
Modal.close(id) sometimes throws the following js error:

rxjs.umd.min.js:50 Uncaught TypeError: Cannot read property 'dispatchEvent' of undefined
    at a (rivet.js:67)
    at Object.p [as close] (rivet.js:1186)
    at ve (util.js:585)
    at a.project (Content.intent.js:85)
    at a._next (rxjs.umd.min.js:128)
...

the offending line is in the fireCustomEvent() method:
     element.dispatchEvent(event)

...when called (in this instance) from Modal.close.

I overcame the issue by simply changing the offending line to:

     element && element.dispatchEvent(event)

Cause of issue:  I’m wondering if perhaps the modal being closed was destroyed (via. something like a state var like showModal  being set to false…and toggling its react rendering) before the close fireEvent was attempted…i.e. a timing issue.

scottanthonymurray:
Fixes #451

### DIFF
--- a/cypress/integration/modal_spec.js
+++ b/cypress/integration/modal_spec.js
@@ -123,4 +123,20 @@ describe('Rivet modal interactions', function() {
 
     cy.get(MODAL).should('have.attr', 'aria-hidden', 'true');
   });
+
+  it('Should not throw an error if removed then opened with .open() method', function() {
+    cy.get(MODAL).then(modal => modal.remove());
+
+    cy.window().then(win => {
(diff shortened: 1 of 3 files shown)


levimcg (approved):
@scottanthonymurray Nice, thanks for fixing this! 💯

PR #470, opened 2021-04-26
Updated loading indicator margin to center XXS indicator in small button

### LINKED ISSUE #391: Loading indicator in small buttons
When I use the smallest loading indicator size (.rvt-loader--xxs) inside of a small button, the alignment seems off:

Screen Shot 2020-05-26 at 10 35 31 AM

When I use the next size up (.rvt-loader--xs), the loading indicator touches the edge borders of the button:

Screen Shot 2020-05-26 at 10 35 55 AM

Is this a positioning bug with the smallest loading indicator size? Or should we always use .rvt-loader--xs regardless of the button size?

scottanthonymurray:
Documentation needs to be updated to say that XXS indicator is the only size you can use in small buttons.

Fixes #391

### DIFF
--- a/src/sass/components/_buttons.scss
+++ b/src/sass/components/_buttons.scss
@@ -302,10 +302,14 @@
     top: 50%;
 
     /**
-         * This negative margin accounts for the width/height of the
-         * default loading indicator above.
-         */
+     * This negative margin accounts for the width/height of the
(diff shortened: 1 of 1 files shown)


levimcg (approved):
This looks good to me. 🚀

PR #474, opened 2021-04-28
Added new Layouts and the new "Summary" layout

levimcg:
I'm going to go ahead and get this PR open. It's the first draft of the first Rivet _Layout_. There's a lot here as I had to create a handful of new components for layouts. ⚠️ These are all in a very alpha state ⚠️ . We'll need to focus on documentation and more testing as we continue building out other layouts.

## New components to support Layouts
- Billboard
- Card
- CTA
- Hero area
- Stat & Stat Group

## New utilities
to support more spacious marketing-type layouts see [spacing tokens](https://github.com/indiana-university/rivet-source/blob/feature/index-summary-layout/src/tokens/spacing.json) for more info:
- `-3-x`
- `4-x`

## General notes
- 🚨  This layout basically uses a lot of new components and some utilities that act as the "glue" to handle vertical spacing between sections. It works, but we may need to think about the possibility of some high-level "layout components" that handle this. I'm not sure?
- I've reworked a lot of components (.njk files) to be more modular and "component-tized" so that they can be combined together in different contexts to build layouts in the future
- The new "Card" component should eventually replace the existing generic "Box" component as well as the the "Feature" component we started last year before the pandemic hit. we can mark them both as deprecated for now and put a note in the changelog/GH release of the next alpha


src/sass/utilities/_spacing.scss
```diff
@@ -27,7 +27,9 @@ $spacing-sizes: (
   -md: $spacing-md,
   -lg: $spacing-lg,
   -xl: $spacing-xl,
-  -xxl: $spacing-xxl
+  -xxl: $spacing-xxl,
+  -3-x: $spacing-3-x,
```
scottanthonymurray, line 31: My expectation from how sizes are written in other contexts (like clothing) was that these classes would be `3xl`, `4xl`, etc.

levimcg: @scottanthonymurray Yeah, that's a good point. I agree. For some reason when I tried that to output was getting translated to `-3-x` and `-4-x` when it's converted from the main `tokens/spacing.json` file to the Sass variables so I opted to match that here in the Sass map that we use to generate all the utility classes. Maybe I'll take a quick look at how were processing the transformation and see if I can get it to work correctly.

levimcg: @scottanthonymurray It looks like due to limitations in the way style dictionary processes property names. The best we can do for the new spacing tokens is `-3-xl` and `4-xl`. Not exactly what I wanted but I think it's workable until we can circle back and think of a better way to handle it.

PR #478, opened 2021-05-05
Updated sidenav API to accept toggle ID rather than element

scottanthonymurray:
You used to have to pass the entire toggle element to `open`/`close` fetched via `querySelector`. This PR changes the sidenav API where you now only have to pass the toggle element ID. This will make client code simpler.

- `open(toggleId)`
- `close(toggleId)`

There's some refactoring we'll need to do in the near future to reduce some duplication in the sidenav source file.

### DIFF
--- a/src/js/components/sidenav.js
+++ b/src/js/components/sidenav.js
@@ -109,7 +109,7 @@ export default class Sidenav extends Component {
           return;
         }
 
-        const openEvent = Component.dispatchCustomEvent('sidenavListOpen', toggle, {
+        const openEvent = Component.dispatchCustomEvent('sidenavListOpened', toggle, {
           id: toggleId
         });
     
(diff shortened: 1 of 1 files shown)


src/js/components/sidenav.js
```diff
@@ -92,30 +92,52 @@ export default class Sidenav extends Component {
...
-        const closeEvent = Component.dispatchCustomEvent('sidenavListClose', toggleButton, {
-          id: toggleButton.dataset.rvtSidenavToggle
+      close(toggleId) {
+        const toggle = this.element.querySelector(
+          `[${this.toggleAttribute}="${toggleId}"]`
+        );
+        const targetList = this.element.querySelector(
+          `[${this.listAttribute}="${toggleId}"]`
+        );
+        if (!toggle || !targetList) {
```
levimcg, line 129: Nitpick, but maybe add an empty line before this `if` statement to match the `open` method above.

levimcg (approved):
Nice, works great! I added one small formatting thing to fix before merging, but that's it.

src/js/components/sidenav.js
scottanthonymurray: Yeah, I didn't bother with any styling/formatting stuff in this PR. My intention is to do a style/refactoring pass in the near future on all the components.

PR #485, opened 2021-05-10
Updated existing Cypress tests to use new wicked-elements API

scottanthonymurray:
This PR updates the existing Cypress tests to use the new `wicked-elements`-based component APIs.

At some point in the near future, we'll need to add some new tests for components like tabs to check programmatic use of the API methods. (For example, the tabs tests currently don't check if programmatically calling `activateTab()` works.)


levimcg (changes requested):
This is looking good to me. I did notice when I ran the test locally that the Modal spec was failing for the "Should be able to close the modal by clicking outside" assertion.

I realized when I opened the Cypress desktop app and stepped through the tests that it's failing because Cypress defaults to clicking right in the center of the viewport which means it's simulating a click inside the modal, which wouldn't close the modal by design. Since the modal is still visible after what we thought was a click on the `body` it fails.

You can fix this by passing either x,y coordinates to the `.cick()` method, or Cypress also provides a shorthand for generic positions on the screen to simulate the click. This would be on line 54 of modal_spec.js

```js
cy.get('body').click('topLeft');
```

I tested it with that ☝🏻 locally and it fixed the failure. [More on the click method in the Cypress docs](https://docs.cypress.io/api/commands/click#Syntax).

PR #487, opened 2021-05-11
Updated accordion and tabs APIs to accept panel ID instead of element

scottanthonymurray:
This PR updates the accordion and tabs APIs to accept a panel ID string instead of the target panel element in `open()`, `close()`, and `activateTab()`. This brings these components in line with how the sidenav API is set up.

I haven't done any code cleanup or refactoring -- this PR just includes the bare minimum to get the new API to function. I'll come back and clean up the component source files in a later PR.

### DIFF
--- a/src/components/tabs/README.md
+++ b/src/components/tabs/README.md
@@ -86,7 +86,7 @@ If you use the appropriate data attribute/id combination in your markup, modals 
 
 | Method                      | Description                                                                                                                      |
 | --------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
-| .activateTab(tab, callback) | - `tab` - The element to be activated. - `callback` - An optional callback function that is executed after the tab is activated. |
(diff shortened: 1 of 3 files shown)


levimcg (approved):
Nice. Looks great. 👍🏻 

I did notice a couple of debugging statements that console.log the id of the accordion tab and and accordion panels. As we think about refactoring we should probably get rid of those as long as they are just debugging statements.

Another thing re: refactoring. I wonder if it makes sense to standardize the way we are naming these two similar methods that open/activate/show-hide an tab or accordion panel? Right now it's `Accordion.open('panel-id')` vs. `Tabs.activateTab('tab-id')` Maybe that's ok? Or, maybe it couuld become `Accordion.activatePanel('panel-id')` to be a bit more consistent between these two very similar (in terms of functionality) widgets. No action needed here, just something that would be worth looking at as we start to refine APIs, etc.

scottanthonymurray:
I agree -- I'll add these notes to Asana so we don't forget. 📋

PR #496, opened 2021-05-27
Fixed issue where non-breaking content cause main content area to overflow

levimcg:
This is a small fix that we've implemented as a patch once we noticed it using `2.0.0-alpha.5` on the v2 docs site.


levimcg:
@scottanthonymurray No real need for a review on this one, just wanted to give you a heads up ☝🏻

PR #504, opened 2021-06-14
Updated component custom events API

scottanthonymurray:
This PR updates the API for component custom events in the following ways:

- Event names are all past tense, i.e. `dropdownOpened`, `modalClosed`
- The value of the `event.target` property is always the emitting component's root DOM element
- The value of the `event.detail` property object only contains additional context about the event when needed, such as which tab was activated. References to relevant DOM elements are passed rather than ID strings.

### DIFF
--- a/src/js/components/accordion.js
+++ b/src/js/components/accordion.js
@@ -196,10 +196,8 @@ export default class Accordion extends Component {
     
         const activationEvent = Component.dispatchCustomEvent(
           'accordionOpened',
-          panel,
-          {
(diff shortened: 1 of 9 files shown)


levimcg (approved):
Nice, this looks great. I pulled it down and tested listening for events, etc. All the cypress tests pass too. 🚀 

@scottanthonymurray would you mind creating an Asana to do to go back and and some cypress tests for listening to custom events. Not super urgent, but something we might want to have in place before the pre-beta clean up/refactor we've been talking about.

PR #507, opened 2021-06-21
Updated check box markup and styles

levimcg:
See discussion thread in #463 for more background on this.

Summary:
- Make updates to markup and styles based on recommendations in #463 
- Removed check, and focus CSS transitions. This has felt clunky to me for a long time now, so I made the decision to move these transitions. I plan to do the same when we make similar updates to the radio inputs discussed in #464 

In the original discussion thread there was mention of also requiring classes on the `input` and `label` elements in addition to the new required `<span class="rvt-checkbox"></span>` wrapper element. Due to some requirements brought up in past issues caused by server-side templating languages that generate hidden fields when their built-in form input helpers are used (E.g. .NET Razr, Laravel Blade, etc. templates) it will need to be a hard requirement that we use element selectors like `.rvt-checkbox input[type=checkbox]"` to scope styles instead of requiring class selectors on children of the `<span class="rvt-checkbox">` wrapper.

@basham I'm not sure why, but it will not let me assign you as a reviewer right now, but I wanted to tag you since you created the original issue. I've basically folded in the code from your [codepen demo](https://codepen.io/basham/pen/yLVmBRJ) and replaced any hard-coded values with variable references.


src/components/checkbox/checkbox.njk
```diff
@@ -1,30 +1,73 @@
...
+        <label for="checkbox-long">
+          <span class="rvt-text-bold">Just a quick note</span><br>
+          <span class="rvt-ts-14 rvt-display-inline-block rvt-color-black-600">This checkbox has a really long label that can wrap on to two lines and still have nice left alignment.</span>
+        </label>
+      </span>
+    </li>
+    <li>
+      <span class="rvt-checkbox">
+        <input type="checkbox" name="checkbox-demo" id="checkbox-hidden-label">
+        <label for="checkbox-hidden-label">
```
basham, line 44: This needs `class="rvt-p-all-none"` to make the hit area just the checkbox (24x24), instead of something bigger (40x24).

Alternatively, we could add a modifier (`.rvt-checkbox--sr-only`?) that could automate this and simplify the markup. See "Option ten" in the demo. This works well in Safari+VoiceOver. And it would work well in the situations where you can't modify the inner markup.

https://codepen.io/basham/pen/yLVmBRJ

```html
<div class="rvt-checkbox rvt-checkbox--sr-only">
  <input type="checkbox" name="checkbox-demo" id="checkbox-10" checked>
  <label for="checkbox-10">Option ten, screen reader only label</label>
</div>
```

```css
.rvt-checkbox--sr-only input[type=checkbox] + label {
  height: 0;
  overflow: hidden;
  padding: 0;
  width: 0;
}
```

src/components/checkbox/checkbox.njk
```diff
@@ -1,30 +1,73 @@
...
+        <input type="checkbox" name="checkbox-demo" id="checkbox-4" disabled checked>
+        <label for="checkbox-4">Option four</label>
+      </span>
+    </li>
+    <li>
+      <span class="rvt-checkbox">
+        <input type="checkbox" name="checkbox-demo" id="checkbox-long">
+        <label for="checkbox-long">
+          <span class="rvt-text-bold">Just a quick note</span><br>
+          <span class="rvt-ts-14 rvt-display-inline-block rvt-color-black-600">This checkbox has a really long label that can wrap on to two lines and still have nice left alignment.</span>
```
basham, line 37: This is a good example, that I hadn't thought of. But hint text like this shouldn't live in the label. Yes, it increases the hit area, but it results in poor experience for screen reader users. That is essentially one long label. In VoiceOver, that gets read out before saying that it is a checkbox and its checked state ("[label] [secondary text], [checked/unchecked], checkbox"). I've done several iterations of this idea (Options 6 to 9 in the [CodePen](https://codepen.io/basham/pen/yLVmBRJ)). The best result is having this hint text separate from the description and attached to the checkbox by `aria-describedby`. This means that the screen reader announces the hint text after it says it is a checkbox ("[label], [checked/unchecked] checkbox, [description]"), which is a better experience.

We could leave it up to the developer to style this (Option 8), but I think we should provide a `.rvt-checkbox__description` to simplify and standardize it (Option 9).

```html
<div class="rvt-checkbox">
  <input aria-describedby="checkbox-9-description" type="checkbox" name="checkbox-demo" id="checkbox-9">
  <label for="checkbox-9">Option nine</label>
  <div id="checkbox-9-description" class="rvt-checkbox__description">This checkbox has a really long label that can wrap on to two lines and still have nice left alignment.</div>
</div>
```

```css
.rvt-checkbox__description {
  color: var(--rvt-color-black-600);
  font-size: calc(14rem/16);
  padding: 0 0.5rem;
}
```

basham (changes requested):
Some of the changes involves removing some properties. If you just copied the CSS from the CodePen, then they could be artifacts from trying to override Rivet v1 styles. So, double check, but I think there are several that would be safe to remove.

After some testing, there are a couple more changes that should be done or discussed, regarding the screen reader and hint text examples.

src/components/checkbox/checkbox.njk
basham: Maybe a better class name is `rvt-checkbox--sr-only-label`? Don't know. Needs workshopped.

src/components/checkbox/checkbox.njk
basham: Also, I'm not convinced bolding the label is needed. The style differences with the description text should be sufficient visual contrast.

src/components/checkbox/checkbox.njk
levimcg: @basham I like the idea of a modifier on the checkbox wrapper element for the uncontrolled markup situations you mention. `rvt-checkbox--sr-only-label` more accurately describes what is going on, but it also feels really verbose. :(

I'm fine with using `rvt-checkbox--sr-only-label`. I'd rather error on the side of clarity.

src/components/checkbox/checkbox.njk
levimcg: @basham Cool, I like option 9 in the demo. I'll make that update and draft some docs for it in the README in Fractal

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,112 @@
...
+  input[type=checkbox]:checked:focus ~ label::before,
+  input[type=checkbox]:indeterminate:focus ~ label::before {
+    box-shadow:
+      0 0 0 ($spacing-xxs / 2) $color-blue-600,
+      0 0 0 $spacing-xxs $color-white-base,
+      0 0 0 ($spacing-xxs * 1.5) $color-blue-600;
+  }
+
+  &__description {
+    color: $color-black-600;
```
basham, line 110: `$color-black-600` passes "normal text" contrast tests for AA but not AAA. If you use `$color-black-700` (`#3d4757`), then that passes both AA and AAA standards. But if you use 700, it's not much contrast against the base color of 800. So, I don't know what to recommend. Just wanted to outline the situation, in case that influences anything. I could be fine with whatever direction.

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,112 @@
...
-    &::before {
-      color: $color-white-base;
-      background-color: $color-blue-600;
-      box-shadow: 0 0 0 .08rem $color-blue-600;
-      background-image: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIxNiIgaGVpZ2h0PSIxNiIgdmlld0JveD0iMCAwIDE2IDE2Ij4KICAgIDxwYXRoIGZpbGw9IiNmZmZmZmYiIGQ9Ik02LjcyLDEzbC0uNDgtLjM2LTMtM0ExLDEsMCwwLDEsNC43MSw4LjI5bDIuMTEsMi4xMiw0LjMzLTYuOTRhMSwxLDAsMCwxLDEuNywxLjA2TDcuNjQsMTIuODdaIi8+Cjwvc3ZnPgo=');
-      background-position: 50% 50%;
-      background-repeat: no-repeat;
-    }
+  &--sr-only-label input[type=checkbox] + label,
+  &--sr-only-label input[type=checkbox] ~ label {
```
basham, line 23: The adjacent sibling connector (`+`) is redundant when also using the general sibling connector (`~`). I'd recommend just using the `~` connector and remove the other, since you're wanting to maintain that option. That would clean several lines of code in this file.

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,112 @@
...
+    overflow: hidden;
+    padding: 0;
+    width: 0;
   }
 
-  /* stylelint-disable */
+  input[type=checkbox] + label,
+  input[type=checkbox] ~ label {
+    display: inline-block;
+    padding: 0 .5rem;
```
basham, line 33: The left padding is needed to put spacing between the checkbox and the label. But the right padding may not be needed. It was there to just maybe put some gap between the right edge of the label and whatever that may be to the right of it, like the edge of a box. But I'm not convinced it is needed. What do you think? This line could be simplified:

```css
padding-left: .5rem;
```

But here's an even further way to simplify things. I originally put padding on the label because of my first attempt with the "checkbox without a visual label" example. But instead, we could remove the padding on the label and the description classes. Then just change the padding on the wrapper.

```css
.rvt-checkbox {
  padding-left: 2rem;
}

.rvt-checkbox--sr-only-label {
  padding-left: 1.5rem;
}
```

With this, if you need padding on the right of the checkbox, then it is easy to add some to the wrapping element.

```html
<div class="rvt-checkbox rvt-p-right-sm">
```

basham (changes requested):
Tested the latest changes in Safari + VoiceOver. Works well.

src/sass/checkboxes/_base.scss
levimcg: As long as we're passing AA, I think we're good here. Of course AAA would be ideal, but given the visual contrast we're trying to achieve. I say leave it as is.

src/sass/checkboxes/_base.scss
levimcg: That's a good call.

src/sass/checkboxes/_base.scss
levimcg: I like the simplified version here. I think it makes a lot more sense.

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,103 @@
...
+  padding-left: $spacing-lg;
+  position: relative;
+
+  input[type=checkbox] {
+    cursor: pointer;
+    height: $spacing-md;
+    left: 0;
+    margin: 0;
+    top: 0;
+    width: $spacing-md;
```
basham, line 17: Now I remember another reason I added left padding on the label. It ensured that the clickable hit area extended fully from the visual checkbox through to the end of the label. Now, there's a gap. But this can be overcome by adjusting the checkbox width. This technique works in Chrome, Firefox, and Safari.

```css
.rvt-checkbox input[type=checkbox] {
  width: 2rem;
}

.rvt-checkbox--sr-only-label input[type=checkbox] {
  width: 1.5rem;
}
```

Despite needing this additional change, I still like doing this, rather than having padding on the label and description.

src/components/checkbox/README.md
```diff
@@ -1,155 +1,80 @@
...
-    checkbox.indeterminate = true;
-</script>
 ```
 
-### Alternate markup with hidden fields
+## Visually-hidden label modifier
+To visually hide the label of a checkbox, but still make it accessible to screen readers, use the `rvt-checkbox--sr-only-label` modifier on the `rvt-checkbox` wrapper element.
 
-This example includes a hidden input in the markup, as some frameworks require. It requires that the input and label be wrapped in a `.rvt-checkbox-wrapper`. Here we are using the wrapper class on a `li`, but it will also work with generic elements like a `div` and `span`.
+The markup in this example will visually hide the label element leaving only the checkbox element visible. This can be useful in situations such as when checkboxes are used to select entire rows of data from a table.
```
basham, line 57: Add a comma after "label element" to help the sentence flow.

src/components/checkbox/README.md
```diff
@@ -1,155 +1,80 @@
...
 
-- Write the label so that it describes the selected state of the checkbox
-- Write the label as a phrase or an imperative sentence, and use no ending punctuation
-        - **Exception:** If a checkbox label also labels a subordinate control that follows it, end the label with a colon
-- Use sentence-style capitalization
-- For a group of checkboxes, use parallel phrasing and try to keep the length about the same for all labels
-- For a group of checkboxes, focus the label text on the differences among the options. If all the options have the same introductory text, move that text to the group label.
-- Use positive phrasing. Don't phrase a label so that selecting a checkbox means not to perform an action.
-        - **Exception: Don't show this again** checkboxes
+## Description Element
```
basham, line 66: To follow the casing elsewhere in this doc, it should be:

```md
## Description element
```

src/components/checkbox/README.md
```diff
@@ -1,155 +1,80 @@
...
-        - Example: Pizza toppings (select one or more)
-
-## External resources
-
-- [Checkboxes vs. Radio Buttons](https://www.nngroup.com/articles/checkboxes-vs-radio-buttons/)
-- [Mozilla Developer Network Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)
-- [MSDN Design Documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742401.aspx)
-- [Indeterminate Checkboxes](https://css-tricks.com/indeterminate-checkboxes/)
\ No newline at end of file
+### Description Element Example
```
basham, line 73: To follow the casing elsewhere in this doc, it should be:

```md
### Description element example
```

src/components/checkbox/README.md
```diff
@@ -1,155 +1,80 @@
...
-- For a group of checkboxes, focus the label text on the differences among the options. If all the options have the same introductory text, move that text to the group label.
-- Use positive phrasing. Don't phrase a label so that selecting a checkbox means not to perform an action.
-        - **Exception: Don't show this again** checkboxes
+## Description Element
+In some cases it can be helpful to provide additional more detailed information about the functionality of a checkbox in addition to the the text inside the `<label>` element. In these instances you can use the `.rvt-checkbox__description` element to provide more context to a user.
 
-### Checkbox group labels
+1. Add an `aria-describedby` attribute with a unique value to the checkbox input
+1. Add a `<div class="rvt-checkbox__description"></div>` element after the <label> element as direct child of the `<span class="rvt-checkbox">` element.
+1. Add an `id` attribute to the description element that matches the `aria-describedby` attribute added in step 1.
```
basham, line 71: This feels like it could be reworked a little. Try something like this:

```md
1. Add a `<div class="rvt-checkbox__description" id="checkbox-description"></div>` element as a direct child of the `<span class="rvt-checkbox">` element, placed after the `<label>` element. Make the id unique for the document.
2. Add an `aria-describedby` attribute to the checkbox `<input>`, referencing the `id` attribute of the description element.
```

This is an alternate version of the second one. Don't know if it is any better.

```md
2. Add an `<input aria-describedby="checkbox-description">` attribute to the checkbox. Reference the `id` attribute of the description element.
```

basham (changes requested):
Getting close!

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,107 @@
...
-  .#{$prefix}-checkbox-wrapper &:disabled ~ .#{$prefix}-label::before {
-    background-color: $color-black-200;
-    box-shadow: 0 0 0 .08rem $color-black-300;
+    &:hover {
+      cursor: pointer;
+    }
   }
 
-  /* stylelint-disable */
+  input[type=checkbox] ~ label::before {
```
basham, line 45: Add a white background to the default checkbox, so that it stands out on non-white backgrounds. This is especially important so that an enabled checkbox doesn't suddenly look disabled if it happens to sit in a section of light gray content.

```scss
background-color: $color-white;
```

src/sass/checkboxes/_base.scss
```diff
@@ -3,121 +3,107 @@
...
+    border-radius: $spacing-xxs / 2;
```
basham, line 46: While it looks fine in Chrome and Safari, Firefox has some odd rendering issues with rounded corners and box-shadow. Some of the background color bleeds through at the corners.

You could resolve this by replacing the box-shadow that fakes a border with a literal `2px` border. In doing this, it may simplify other parts in this file. But you will need to shift the left and top position by `2px` to put it back in the same visual position as it currently is.

If this is too nit-picky, that's fine. Just trying to solve problems as I see them, no matter how minor.

basham (changes requested):

src/sass/checkboxes/_base.scss
basham: I'm also fine with doing this work in another branch and merging it in, if you don't want to do it. It may give me good practice with contributing to this repo.

levimcg: That's actually a good idea. I'm fine with tackling that as a separate issue later.

basham (approved):

PR #508, opened 2021-06-23
Fix checkbox rendering in Firefox

basham:
Rounded corners in Firefox causes background colors to bleed through at the corners, when using box-shadow. Replacing the box-shadow border with a literal border fixes this. Works in Chrome, Safari, and Firefox. Also, did some other cleanup.

### DIFF
--- a/src/sass/checkboxes/_base.scss
+++ b/src/sass/checkboxes/_base.scss
@@ -35,11 +35,8 @@
   }
 
   input[type=checkbox] ~ label {
+    cursor: pointer;
     display: inline-block;
-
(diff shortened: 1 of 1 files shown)


basham:
Tagging @levimcg, because I can't add reviewers.

levimcg (approved):
@basham This looks great. I really like the updated sharper corners instead of the rounded corners in this case. I think it will really help the checkboxes stand out against radio button in forms that might have both. Bonus UX win in addition to fixing the FF rendering issue 👍🏻

PR #510, opened 2021-06-30
Updated radio markup and styles

basham:
See discussion thread in #464 for more background on this.

Summary:

1. Make updates to markup and styles, based on recommendations in #464.
1. Removed focus CSS transitions.
1. Added screen-reader-only label styling option.
1. Added radio description option.
1. Removed need for `rvt-radio-wrapper` and `rvt-label` classes.
1. Repurposed `rvt-radio` class to be on a wrapping element, instead on the input element.


basham:
Ready for review, @levimcg.

src/sass/radios/_base.scss
```diff
@@ -3,104 +3,93 @@
 
 @use '../core' as *;
 
+$padding: 3rem / 16;
```
levimcg, line 6: I don't have any real objection to using this sort of "local" variable here to reference in this component specifically. Generally what we've done in the past is create and one-off or "magic" numbers using the built-in `$spacing-*` variables in an in-line math expression. Like in this case you could use `$spacing-sm` which equals 1rem. So something like `($spacing-sm * 3)` to get the `3rem` output you're looking for.

Like I said, I honestly don't have a problem with this and don't really see it as any kind of maintenance issue, so I'll leave that up to you. 👍🏻

levimcg (approved):
@basham One little nitpick here, but otherwise this looks great. Really nice work. I really appreciate you drafting the docs too. 🙌🏻  🚀

src/sass/radios/_base.scss
levimcg: @basham just glanced at this after I submitted the PR review and realized that I totally read it wrong and that you're weren't looking for a `3rem` output here 🙃 . Totally feel free to disregard this comment.

basham: I'm fine making adjustments to fit the code style you've established. But yes, abstracting away the base value would be good.

1. Is it fine having the variable in the top-level of the file, as it is now?
2. Is "padding" a good name?
3. Should the value be this, instead? `$spacing-sm * 3 / 16`

levimcg: @basham The more I think about it, it is totally fine to leave the variable as is. We are also using the latest version of Sass in the project now which means that this local `$padding` variable is technically scoped to this "module". The only other concern I had was that if someone redeclared a variable called `$padding` down the road that here could be an issue, but with Sass modules you would have to intentionally import and then reference the `$padding` variable in another module like:

```scss
@use './path/to/radio/base' as radios;`

// And then use it like this
.some-class {
  padding: radios.$padding;
}
```

PR #511, opened 2021-06-30
Enforce checkbox block styling in CSS, HTML, and examples

basham:
Summary:

1. Replace `rvt-plain-list rvt-flex` example with `rvt-inline-list`.
2. Enforce line-height for checkbox labels, so the effective height is 24px.
3. Enforce the description element to be styled as a block, in case the markup is not a `<div>`.
4. Update all examples to use `<div>` over `<span>`.


basham:
Ready for review, @levimcg.

levimcg (approved):
Looks great!

PR #512, opened 2021-06-30
Use flex for inline lists

basham:
Notes:

1. The inline list implementation is replaced with flex.
2. List items no longer set a line height or font size.
3. Inline list item margins are reset only on direct children, in case there would be some odd case in which you have nested lists under an inline list.
4. Spacing is managed by the [gap property](https://developer.mozilla.org/en-US/docs/Web/CSS/gap), which has [good browser support](https://caniuse.com/flexbox-gap), outside of IE.
5. Top margin is included on inline lists to maintain current spacing expectations (top margins on list items). This could be adjusted, if such spacing is reviewed for all list styles.
6. Removed the Sass mixin, to keep the code simpler and remove unintended classes (`.plain-list`, `.inline-list`).

### DIFF
--- a/src/sass/lists/_base.scss
+++ b/src/sass/lists/_base.scss
@@ -30,33 +30,20 @@ dl dd {
   margin-left: 0;
 }
 
-@mixin plain-list {
+.#{$prefix}-plain-list {
   list-style: none;
   padding: 0;
 }
 
-@mixin inline-list {
-  /**
-   * This is a hack that remove extras space between elements that are
-   * set to display inline-block.
-   */
-  font-size: 0;
-
-  @include plain-list;
-
-  & li {
-    display: inline-block;
-    line-height: 1;
-    font-size: 1rem;
-    margin-right: $spacing-sm;
-    margin-bottom: $spacing-xs;
-  }
-}
-
-.#{$prefix}-plain-list {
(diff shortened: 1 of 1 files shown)


basham:
Ready for review, @levimcg.

levimcg (approved):
@basham Really nice subtle improvement. Love it 👍🏻 

This is a whole new can of worms, but something that it seems like might be another nice small-ish project, but one thing I've been thinking about is a few inconsistencies in naming conventions on a handful of components and utilities.

In some places we are using a format like; `thing-variant` or `thing--variant` (when BEM is involved. Other places its `variant-thing`. For example, we use `rvt-plain-list` here but in the link utils we use `rvt-link-bold`.

There may be a few other places this is happening, but I can't think of them off the top of my head. Might be something to look into fixing in future PRs.

basham:
If you think it would be better, the styles could be abstracted a bit more like this, to reduce repeated properties in the inline list

```scss
.#{$prefix}-plain-list,
.#{$prefix}-inline-list {
  list-style: none;
  padding: 0;
}
```

levimcg:
Oh, yeah, I didn't catch that at all. That would definitely be a nice optimization. Not super critical, IMO.

PR #513, opened 2021-07-01
Update lists

basham:
Summary:

1. `rvt-plain-list` is renamed to `rvt-list-plain`.
2. `rvt-inline-list` is renamed to `rvt-list-inline`.
3. Instead of placing top margins on list items, spacing for all list types are determined by flex gap.
4. Since all list styles are configured by a single class name (no child or descendant selectors against elements), specificity is low, so it is easy to override settings as needed.
5. Top margin is kept at `0.5rem`, to keep good spacing against legend elements.
6. Left padding for ordered/unordered lists is reduced from 40px to 32px.
7. Added nested list examples for `rvt-list` and `rvt-list-plain`.
8. `ol`, `ul`, and `li` elements are no longer directly styled by Rivet. A list class must explicitly be applied.
9. Plain and inline lists retain their list semantics in VoiceOver by using `list-style: ''` instead of `list-style: none`.


basham:
Ready for review, @levimcg.

basham:
For more details on the VoiceOver bug/feature that requires us to use `list-style: ''` instead of `list-style: none`.
https://gerardkcohen.me/writing/2017/voiceover-list-style-type.html

levimcg (approved):
Looks great, thanks!

basham:
More discussion regarding VoiceOver's behavior regarding `list-style: none`:

- https://css-tricks.com/web-features-that-may-not-work-as-youd-expect/#screen-readers-might-not-relay-the-semantics-of-certain-elements
- https://yatil.net/posts/2019/01/13/much-ado-about-no-lists/

PR #514, opened 2021-07-09
Update input background color to white

basham:
Updating the input background color to white ensures that inputs retain good contrast against contextual color. Also, if an input is placed on a light gray background, it makes it so the inputs don't accidentally look disabled.


basham:
Ready for review, @levimcg.

levimcg (approved):
@basham Looks great. Nice update.

PR #515, opened 2021-07-09
Update description list

basham:
## Summary:
1. Adopting [terminology on MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl): `dl` is "description list". `dt` is "description term". `dd` is "description details".
2. Replaced lorem-ipsum in examples with plant examples.
3. Examples show how these solutions work when `div` tags wrap content in a `dl`.
4. Lots of code will be cleaned up and the readme updated once the following questions are answered.

## Questions:
1. How do we accommodate specificity? `list-description-1` uses BEM elements to maintain low specificity. But this means more class names. `list-description-2` uses [`:where()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) to maintain zero specificity. But this is not compatible with Internet Explorer. `list-description-3` uses element selectors, making it more difficult to override styles. What direction should we go? Are there other options? Is there a way to be forward like 2, yet backwards compatible like 1?
2. Is it odd to use top margins here for spacing, but gap for spacing in the vertical lists of `rvt-list` and `rvt-list-plain`?
3. There is `0.5rem` gap between items in the other list types. Should there be something like that for these? If so, should there be a gapless variant? A gapless variant seems useful for cases in which there is 1-term-to-1-details. But if there are more details per term or if the details span multiple lines, having the extra gap helps to distinguish the details from each other.


basham:
Ready for discussion, @levimcg.

levimcg:
@basham Just getting caught up from some PTO. I should be able to get this by the end of day tomorrow/Friday 07/16.

basham:
This article has a number of examples of `dl`s.
https://benmyers.dev/blog/on-the-dl/

scottanthonymurray:
Work on Rivet 2 has diverged significantly since these PRs were first created. Closing for now, but we can definitely revisit in the future.

basham:
This PR was against Rivet 2. As far as I can tell, this branch could be merged with little to no more rework. Please explain how this PR is no longer relevant.

levimcg:
@basham sorry, that's my bad. I asked Scott to close these, but forgot the specifics of this one before I looked at it. I'll re-open this and maybe you and I can dig into wrapping it up here in the next couple of weeks?

levimcg:
@basham Wanted to pick this up again and respond to your original questions from above.

1. I like the use of `:where()` in the `-description-2` examples to deal with specificity issues. I'm fine not supporting IE here, We've started using `:where()` and other non-IE supported CSS features elsewhere in the codebase as progressive enhancements. The fallback experience is will just be the default UA styles and that's totally fine. I'd like to stay away from the BEM class names on details and terms similar to how we're doing on the other list types.
2. I don't have a strong opinion either way here. Similar to `:where()` above, we've started using gap elsewhere and as you say we're already using it in the other list types. The only nice thing about using margin is that you can use utility classes to adjust or remove margins on individual items. We don't have that with gap and I don't feel like adding some sort of gap utilities makes a lot of sense right now.
3. I think part of my answer to question 2 applies here as well (re: utilities), but I also don't have a super strong opinion.

basham:
I've made the changes as we discussed yesterday. Some notes:

1. I wrapped the element selectors (dt, dd) in a `:where()` rather than the `rvt-list-description` class. This makes it more consistent with how other classes are defined.
2. Thanks for the [example of using `@extend`](https://codepen.io/levimcg/pen/OJvQwGx). I was able to extend the class itself, rather than from a mixin. That made the code cleaner in `lists/_base.scss`.
3. I added an example with `rvt-prose`.
4. There are three examples now in `list-description.njk`. Do we need them all, or does it matter?

levimcg (approved):
@basham Nice! This looks great. Everything is working as expected for me 👍🏻. I'm ready to merge this. I will create a task to get the documentation site update to reflect these changes on for the 2.0.0 release.

Following up on your latest questions/comments:

1. Great strategy. I'd like to re-evaluate how we are writing our selectors inside of the `.rvt-prose` utility and see how we might apply this strategy to the other selectors and the additional list types.
2. It's nice to just be able to import the `../lists` and reuse those styles 💯 without duplicating rule sets
3. Looks great!
4. I think it's fine to leave these three examples here for any future testing. I think we're going to look at moving off fractal at some point post-2.0.0 as it is currently [not being maintained/looking for maintainers](https://github.com/frctl/fractal/issues/1167) (bummer). We'll probably clean a lot of things up when we make those updates.

PR #516, opened 2021-07-13
Update typescale with line-height

basham:
## Summary:

1. Typescale now includes line-height, in addition to font-size. This ensures a more consistent vertical rhythm, based on a `4px` grid.
2. Sass typescale variables are replaced in favor of explicit `$font-size-` and `$line-height-` variables.
3. Scales are no longer based on pixel size or t-shirt sizing, as this is difficult for users to remember. Instead, the scale is simply linear, 1 to 13.
4. Font sizes and line heights that are meant to pair with each other share the same scale number.
5. Use `rvt-fs-*` to adjust just the font size.
6. Use `rvt-lh-*` to adjust just the line height.
7. Use `rvt-ts-*` to adjust both the font-size and the line-height.
8. All the Rivet components should be reviewed to ensure they align to the 4px grid. This will likely involve adding more explicit line-heights to selectors.
8. This work is based on this [CodePen proposal](https://codepen.io/basham/pen/oNBVdVg).

## Questions:

1. Is this a direction we want to go?
2. Should font-size and line-height classes be abbreviated like `rvt-fs` and `rvt-lh`?
3. Is a linear scale a better replacement than based on pixels and t-shirt sizing?
4. Do the font-size and line-height classes need breakpoint variations, like the typescale class?
5. Are the suggested line-heights fine for the font-size, or do they need further refinement?


basham:
Ready for discussion, @levimcg.

levimcg:
@basham Same here, just getting caught up from some PTO. I should be able to get this by the end of day tomorrow/Friday 07/16

scottanthonymurray:
Work on Rivet 2 has diverged significantly since these PRs were first created. Closing for now, but we can definitely revisit in the future.

basham:
Could you please explain what specifically has diverged in Rivet 2 that would make this proposal no longer relevant? If you explain that, I can update the proposal for Rivet 2.

basham:
Even if the numbers-based naming convention is not adopted, I still strongly feel that changes to line height are warranted (be it on individual components or utility classes), to align content on a 4px grid.

basham:
I'm fine making such contributions before the 2.0 release, as long as we have a plan for reviewing and adopting the work.

PR #518, opened 2021-08-10
Removed up and down arrow key listeners from tabs component

### LINKED ISSUE #509: Rivet Tabs breaks keyboard control of browser scrolling when focus is in the tab list
## Describe the bug
The Rivet Tab widget responds to the up/down arrow keys and uses them to move focus from tab to tab when tabs are arranged horizontally. This behavior does not conform to the ARIA spec. Screen-reader users are not able to navigate the page using the up/down arrows (verified with JAWS)

## Expected behavior
When focus is on a tab and the tabs are arranged horizontally, the widget should not listen for and respond to the up/down arrow keys. 

- See example Tab widget at: https://rivet.iu.edu/components/page-content/tabs/ 
- See Note # 3 on: https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel : 
> 3. If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

## Steps to reproduce
Add steps to reproduce the bug.

1. Using Chrome Version 91.0.4472.101 (Official Build) (64-bit) on Windows 10 with JAWS 2021.2105.53 ILM
2. Navigate to: https://rivet.iu.edu/components/page-content/tabs/ 
3. Move focus to one of the tabs in either of the example IU Rivet Tab widgets on the page.
4. Press up/down arrows and note that the active tab changes and JAWS only speaks the newly active tab's accessible name instead of page content above/below as indicated by the arrow key pressed.

scottanthonymurray:
Fixes #509

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "rivet-uits",
-  "version": "1.8.0",
+  "version": "1.8.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/js/components/tabs.js
+++ b/src/js/components/tabs.js
@@ -176,25 +176,13 @@ var Tabs = (function() {
     var prevTab = tabs.indexOf(activeTab) - 1;
 
     switch (event.keyCode) {
-      case KEYS.right || KEYS.down:
-        !tabs[nextTab] ?
-          tabs[0].focus() :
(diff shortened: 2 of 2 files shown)


levimcg (approved):
Nice, looks great. 💯

PR #528, opened 2021-09-22
Updated colors to align with new IU brand palette

levimcg:
This PR brings the design tokens and color palette in sync wit the recently updated IU [brand color palette](https://www.iu.edu/brand/brand-expression/visual-language/color/index.html).

@scottanthonymurray I've transposed colors from the old palette to the new and I think everything is good, but if you could pull this down and do a quick visual inspection just to make sure, that would be great.


scottanthonymurray (approved):
@levimcg Looked through all the components in Fractal and didn't notice anything obviously busted. 🏄‍♀️

PR #558, opened 2021-12-28
Refactored component JavaScript

scottanthonymurray:
To-dos:

- [ ] Fix failing Cypress tests


cypress/integration/dropdown_spec.js
```diff
@@ -69,8 +69,8 @@ describe('Dropdown Interaction', function() {
     cy.get(DROPDOWN_TOGGLE).should('have.attr', 'aria-expanded', 'false');
```
levimcg, line 70: As discussed yesterday face-to-face, we should add additional tests in here somewhere to test to make sure it is closing when clicking outside of the menu.

src/js/components/dropdown.js
```diff
@@ -1,291 +1,395 @@
...
+        return dispatched
+      },
+
+      /************************************************************************
+       * Handles click events broadcast to the disclosure.
+       *
+       * @param {Event} event - Click event
+       ***********************************************************************/
+
+      onClick (event) {
```
levimcg, line 226: We discussed this face-to-face, but we need to make sure we add back in the functionality that closes the dropdown menu when the click came from outside the dropdown menu itself. So if you say click on the document body, or another adjacent dropdown toggle (like in the header) it should close the currently open dropdown.

levimcg (changes requested):
I think this is almost there. I left a few inline comments and there are a couple more things I noticed I'll list here in this wrap-up.

1. You missed a few references to the `vendor.js` file that are still in some of the gulp/fractal related files that is breaking the build process. There are references in these files that need removed: `gulpfile.js` (a task called `vendorJS`), `gulp/javaScriptTasks.js`, and `fractal.config.js`
2. It looks like you updated the copyright date to 2022, but since the majority of these files were first open-sourced in 2018, that original copyright should not change. Event though it has changed and been refactored, etc, the copyright has to remain the same as the first year it was published per the IU [open source initiative guidelines](https://indiana-university.github.io/). It looks like you created a new file in polyfills, that new file should have a 2022 copyright date.

levimcg (approved):
Nice, work. Everything looks good to me 🚀

PR #564, opened 2022-01-28
Updated Sass to use math.div()

scottanthonymurray:
The `/` operator for division is deprecated.

See https://app.asana.com/0/1199946964816580/1200802859946356/f

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13398,13 +13398,14 @@
       "dev": true
     },
     "sass": {
-      "version": "1.44.0",
-      "resolved": "https://registry.npmjs.org/sass/-/sass-1.44.0.tgz",
-      "integrity": "sha512-0hLREbHFXGQqls/K8X+koeP+ogFRPF4ZqetVB19b7Cst9Er8cOR0rc6RU7MaI4W1JmUShd1BPgPoeqmmgMMYFw==",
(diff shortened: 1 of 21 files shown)


scottanthonymurray:
@levimcg Some division expressions had enclosing parentheses around them presumably for syntax or order-of-operations reasons. I left them as they were, but I think the use of `math.div()` makes a lot of them redundant so let me know if you'd like me to remove them as well.

levimcg (changes requested):
@scottanthonymurray We'll actually have to update to at least `sass@1.33.0` before we can make this update. Right now the sass task errors out because the `.div` function is not available in the version (`1.25.0`). Here's the docs on that: https://sass-lang.com/documentation/modules/math#other-functions

That's my bad. I had written up [a separate task](https://app.asana.com/0/1199946964816580/1201366217856409/f) for updating to the latest sass and should have mentioned that in this task.

levimcg (commented):
@scottanthonymurray Great, thanks for updating that dependency.

RE: the extra parenthesis, if you could go ahead and removed those, that would be awesome. You're right, with the new math.div function it makes them unnecessary. They were mostly there for readability reasons.

levimcg (approved):
Nice, thank you!

PR #570, opened 2022-02-03
Added Benton Sans Medium font-weight

levimcg:
This PR adds the medium weight of Benton Sans to align with updated branding guidelines.

### DIFF
--- a/src/components/01-utilities/text.njk
+++ b/src/components/01-utilities/text.njk
@@ -1,6 +1,7 @@
 <h1 class="rvt-ts-32 m-bottom-lg">Text utilities</h1>
 
 <p class="rvt-text-regular">Regular text utility</p>
+<p class="rvt-text-medium">Medium text utility (Benton Sans only)</p>
 <p class="rvt-text-bold">Bold text utility</p>
 <p class="rvt-text-uppercase">Uppercase text utility</p>
 <p class="rvt-text-left">Left aligned text utility</p>
--- a/src/sass/defaults/_fonts.scss
+++ b/src/sass/defaults/_fonts.scss
@@ -19,6 +19,19 @@
 @font-face {
   font-family: BentonSans;
   font-style: normal;
+  font-weight: 600;
+  src: url('https://fonts.iu.edu/fonts/benton-sans-medium.eot');
+  src:
+    url('https://fonts.iu.edu/fonts/benton-sans-medium.eot?#iefix') format('embedded-opentype'),
+    url('https://fonts.iu.edu/fonts/benton-sans-medium.woff') format('woff'),
+    url('https://fonts.iu.edu/fonts/benton-sans-medium.ttf') format('truetype'),
+    url('https://fonts.iu.edu/fonts/benton-sans-medium.svg#BentonSansMedium') format('svg');
+  font-display: swap;
+}
+
+@font-face {
+  font-family: BentonSans;
+  font-style: normal;
   font-weight: $font-weight-bold;
   src: url('https://fonts.iu.edu/fonts/benton-sans-bold.eot');
   src:
--- a/src/sass/defaults/_fonts.scss
+++ b/src/sass/defaults/_fonts.scss
@@ -19,7 +19,7 @@
 @font-face {
   font-family: BentonSans;
   font-style: normal;
-  font-weight: 600;
+  font-weight: $font-weight-medium;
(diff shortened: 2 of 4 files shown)


scottanthonymurray:
@levimcg What are your thoughts on referring to Benton Sans Medium as semibold instead of medium and using `.rvt-text-semibold` instead of `.rvt-text-medium`?

According to MDN/OpenType, weight 600 is typically "semibold" whereas 500 is "medium". My gut reaction was also that "medium" and "regular" are somewhat synonymous and might be confusing for developers. I might be in the minority in this thinking though.

levimcg:
@scottanthonymurray Yeah, that totally makes sense. I think there is a bit of nuance in this situation that makes it a bit more complicated.

Benton Sans doesn't technically have a Semibold weight so there could be some confusion for example, in desktop design applications like the Adobe suite, Figma, etc. When a designer imports the desktop fonts those will be named "Benton Sans Medium" because the names in the UI correspond to the font files. So we would end up with a mismatch between what we call the font weight in developer land vs. designer land because there is technically no font called "Benton Sans Semibold" in the family (desktop or web fonts).

All that said the we can definitely alias the medium weight to a value of `500` for the `font-weight` CSS property. We set those values in the `_fonts.scss` partial, so that number value that corresponds to the "Benton Sans Medium" font files can technically be whatever we want.

Does that makes sense? You bring up a good point about the ambiguity between regular and medium. I wonder if we should be more explicit about how we name the utility classes/sass variables. Eg. `$font-weight-500` -> `.rvt-font-weight-500` instead of `.rvt-text-medium`.

scottanthonymurray:
@levimcg I'm split on renaming the utilities. I think it's easier to comprehend and remember the abstract weights (regular, bold, etc) rather than the specific numbers, but maybe it won't be an issue for most developers.

levimcg:
@scottanthonymurray 

> I think it's easier to comprehend and remember the abstract weights (regular, bold, etc) rather than the specific numbers

Totally agree 👍🏻. I've updated the actual font-weight token to use 500 for the weight since that is the standard mapping for "Medium" font weights.

I'm fine moving forward with `rvt-text-medium` for the css class, but we could also try adding the alias class of `.rvt-font-weight-medium` along with `.rvt-text-medium` to see if that would clear things up a bit. a `font-weight-*` naming convention maps better to the design token name of `$font-weight-*`. I dunno, thoughts?

scottanthonymurray:
@levimcg Just so I'm clear, we would have:

- `.rvt-font-weight-regular`
- `.rvt-font-weight-medium`
- `.rvt-font-weight-bold`

And that we would alias these to

- `.rvt-text-regular`
- `.rvt-text-medium`
- `.rvt-text-bold`

If so, I think we should deprecate the `.rvt-text-*` classes listed above. This might potentially introduce a consistency issue with other text utility classes, many of which follow the `.rvt-text-*` naming convention (for example, we don't have `.rvt-text-align-center`, we have `.rvt-text-center`).

It's a minor thing but I'd like to avoid situations where a developer has to remember which prefix applies to which classes in the same family of utilities. Maybe we can split the difference and do something like `.rvt-text-weight-500`, etc?

levimcg:
@scottanthonymurray Oooh, I like that, but I say we definitely go with the "regular, medium, bold" language like you recommended instead of the number values.

How about this?
- `.rvt-text-weight-regular`
- `.rvt-text-weight-medium`
- `.rvt-text-weight-bold`

And we deprecate the old `.rvt-text-regular`, etc. versions. By deprecat, I mean that we'll alias `.rvt-text-weight-regular` to `.rvt-text-regular` and note that in the upcoming release. Then remove the alias at the time of 2.0.0 production.

scottanthonymurray:
@levimcg That sounds good to me! 🙌

scottanthonymurray:
@levimcg Actually, sorry, now that I think about it, is there really a meaningful difference from a developer comprehension perspective between something like `.rvt-text-weight-regular` and `.rvt-text-regular`? I wonder now if maybe I've overthought it and it's sufficient to just have

- `.rvt-text-regular`
- `.rvt-text-medium`
- `.rvt-text-bold`

And explain clearly in the documentation that these terms map to those used by CSS to refer to the various numeric weights (400, 500, and 700)?

levimcg:
@scottanthonymurray That sounds good to me. If we're seeing that there is confusion among developers, we can adjust.

PR #575, opened 2022-02-17
Reduced font size of small button variant

scottanthonymurray:
https://app.asana.com/0/1199946964816580/1201710369645723/f

### DIFF
--- a/src/sass/buttons/_base.scss
+++ b/src/sass/buttons/_base.scss
@@ -165,11 +165,13 @@
 }
 
 .#{$prefix}-button--small {
+  font-size: $ts-14;
   height: $spacing-md;
   line-height: 1.1;
   vertical-align: bottom;
   padding-right: $spacing-xs;
   padding-left: $spacing-xs;
+  padding-top: $spacing-xxs * 0.5;
 }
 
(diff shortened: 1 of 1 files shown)


src/sass/buttons/_base.scss
```diff
@@ -165,11 +165,13 @@
 }
 
 .#{$prefix}-button--small {
+  font-size: $ts-14;
   height: $spacing-md;
   line-height: 1.1;
   vertical-align: bottom;
   padding-right: $spacing-xs;
   padding-left: $spacing-xs;
+  padding-top: $spacing-xxs * 0.5;
```
levimcg, line 174: I think you're right in the particular example we have in the fractal site with the text being "Primary" this padding does help make the text appear more vertically centered optically. But, I worry that with different words that have various combinations of ascenders and descenders it could have the opposite affect making it look like there's more space on top than bottom.

See this screenshot below. The example on the right with a combination of uppercase letters and those with descenders ("y" and "g") look off center with the small amount of top padding applied.

image

levimcg (changes requested):

levimcg (approved):
Nice, thanks for making that change!

PR #577, opened 2022-02-23
Updated secondary button styles to increase contrast between text and border

scottanthonymurray:


levimcg (approved):
Nice, looks great. 💯

PR #578, opened 2022-02-23
Updated accordion styles to match similar styles used by link hub

scottanthonymurray:
This PR updates the accordion component styles to be more consistent with similar styles used by the link hub component.

### DIFF
--- a/src/components/accordion/accordion.njk
+++ b/src/components/accordion/accordion.njk
@@ -4,7 +4,7 @@
       <button class="rvt-accordion__toggle" id="{{ id }}-{{ loop.index }}-label" data-rvt-accordion-trigger="{{ id }}-{{ loop.index }}">
         <span class="rvt-accordion__toggle-text">{{ panel.summary }}</span>
         <div class="rvt-accordion__toggle-icon">
-          <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 16 16">
+          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
             <g fill="currentColor">
               <path class="rvt-accordion__icon-bar" d="M8,15a1,1,0,0,1-1-1V2A1,1,0,0,1,9,2V14A1,1,0,0,1,8,15Z"/>
               <path d="M14,9H2A1,1,0,0,1,2,7H14a1,1,0,0,1,0,2Z"/>
--- a/src/components/accordion/accordion.njk
+++ b/src/components/accordion/accordion.njk
@@ -4,7 +4,7 @@
       <button class="rvt-accordion__toggle" id="{{ id }}-{{ loop.index }}-label" data-rvt-accordion-trigger="{{ id }}-{{ loop.index }}">
         <span class="rvt-accordion__toggle-text">{{ panel.summary }}</span>
         <div class="rvt-accordion__toggle-icon">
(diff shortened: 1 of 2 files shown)


src/components/accordion/accordion.njk
```diff
@@ -4,7 +4,7 @@
       <button class="rvt-accordion__toggle" id="{{ id }}-{{ loop.index }}-label" data-rvt-accordion-trigger="{{ id }}-{{ loop.index }}">
         <span class="rvt-accordion__toggle-text">{{ panel.summary }}</span>
         <div class="rvt-accordion__toggle-icon">
-          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
+          <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 16 16">
```
levimcg, line 7: I think we might need to handle increasing the size of this icon as a separate issue/PR. For whatever reason increasing the size to 20px is causing some "fuzziness" in the rendering. There may just be some jank in the source svg icon. It looks good at 16px which is the size the icons was drawn at, but for some reason at 20px it's not rendering sharply like it should. Screenshots below show the original 16px svg on the top vs. the upsized 20px svg on the bottom. The rendering is pretty fuzzy.

Let's set the `width` and `height` attributes back to `16` for now and I can create a separate issue/task to address this. We may just need to create a custom icon that is dialed in to work at this size, or redraw the plus icon from the rivet icons set.

image

image

src/sass/accordion/_base.scss
```diff
@@ -38,18 +38,32 @@
...
+    font-size: $ts-20;
+    font-weight: $font-weight-regular;
     line-height: 1.3;
     padding-right: $spacing-sm;
   }
 
   &__toggle-icon {
     color: $color-black-400;
     margin-left: auto;
+    position: relative;
```
levimcg, line 50: Instead of messing with the positioning here, we can make this a little more flexible and resilient by setting the actual icon element to display flex and align-items to center. That way no matter what the icon size is, it will be vertically centered in relation to the toggle text.

```diff
- position: relative;
- top: 4px;
+ display: flex;
+ align-items: center;
```

src/sass/accordion/_base.scss
```diff
@@ -38,18 +38,32 @@
...
+    top: 4px;
   }
 
   &__panel {
-    padding: $spacing-md;
+    padding: $spacing-xs $spacing-sm $spacing-md $spacing-sm;
+  }
+
+  &__panel > * {
+    margin-top: 0;
```
levimcg, line 59: Instead of creating an extra rule set below with the `*:last-child block`, we can zero out the bottom margin here which is what we do in the `.rvt-prose` utility.

The point here being that we only apply margin/spacing to the _top_ of direct child elements to keep elements that might have both top and bottom margins from collapsing, causing inconsistent spacing. 

```diff
&__panel > * {
  margin-top: 0;
+ margin-bottom: 0;
}
```

src/sass/accordion/_base.scss
```diff
@@ -38,18 +38,32 @@
...
+  }
+
+  &__panel > * + * {
+    margin-top: 1.5rem;
+  }
+
+  &__panel > *:last-child {
```
levimcg, line 66: After removing the bottom margin for all elements in the block above, this can be deleted.

levimcg (changes requested):
Couple of comments/changes here, but otherwise great job on translating the new design. 💯

levimcg (approved):
💯

PR #580, opened 2022-03-01
Added new CTA button variant

scottanthonymurray:


levimcg (commented):
@scottanthonymurray This looks great so far. I took a look and wrote down some notes that I'd like to discuss 1:1 (maybe tomorrow when we meet). I want to make sure we're getting the ergonomics right from a developer standpoint and it might be easer to has a few things out face to face and then I can add my review here once we move it from draft status.

levimcg:
@scottanthonymurray I finished this up and ended up making some design changes that carry over the arrow icon from the default cta. I thought it made more sense here to follow BEM conventions and make the new `rvt-cta--button` variant, since there is more commonality between the two variants. We can discuss when we document, etc.

PR #586, opened 2022-04-04
Updated beta workflow node version to 14

scottanthonymurray:
Updating since Node.js version 12 reaches end of life on April 30, 2022.


levimcg (approved):
Nice, thanks! 💯

PR #588, opened 2022-04-13
Added dark background hero variant

levimcg:
Added initial support for dark background. We'll need to add and example of this to the docs in the 2.0.0-beta.4 release.


levimcg:
**Updated 05-03-22**: See original PR #587 for original updates. This added support for dark backgrounds with white text after some initial testing.

PR #603, opened 2022-06-30
Updated build process to use rimraf

scottanthonymurray:
This PR updates the build process to use the platform-agnostic `rimraf` command instead of `rm -rf`.

You'll need to run `npm i` to install the `rimraf` package before running the build process.


levimcg (approved):
Nice, works great!

PR #604, opened 2022-06-30
Added line-height tokens and utility classes

scottanthonymurray:
This PR adds the following line-height tokens and utility classes:

- `$line-height-tight` / `.rvt-lh-tight`: `1.2`
- `$line-height-base` / `.rvt-lh-base`: `1.5`
- `$line-height-loose` / `.rvt-lh-loose`: `1.65`

The `.rvt-lh-title` class remains for backwards compatibility.

### DIFF
--- a/src/tokens/line-heights.json
+++ b/src/tokens/line-heights.json
@@ -1,0 +1,13 @@
+{
+  "line-height": {
+    "base": {
+      "value": "1.5"
+    },
+    "heading": {
+      "value": "1.2"
+    },
+    "taller": {
+      "value": "1.65"
+    }
+  }
+}
--- a/src/tokens/line-heights.json
+++ b/src/tokens/line-heights.json
@@ -6,7 +6,10 @@
     "title": {
       "value": "1.2"
     },
-    "taller": {
+    "tight": {
+      "value": "1.2"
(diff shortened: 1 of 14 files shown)


src/tokens/line-heights.json
```diff
@@ -0,0 +1,16 @@
+{
+  "line-height": {
+    "base": {
+      "value": "1.5"
+    },
+    "title": {
```
levimcg, line 6: Since these `title` and `tight` are the same value, can we ditch `title` all together and replace all instances where we're using `title` with `tight`?

I think it will be more straightforward to just have the three values for now instead of duplicating `tight` and `title` since `title` doesn't necessarily match the naming convention.

scottanthonymurray: Agreed! Thanks :)

scottanthonymurray: @levimcg Alright, I've removed the `.rvt-lh-title` class. ✂️

levimcg: Awesome. I think we can go ahead an remove it from the `line-heights.json` file as well and we'll be good to go!

levimcg (approved):

PR #607, opened 2022-07-05
Implemented data-rvt-disclosure-open-on-init attribute

scottanthonymurray:
This PR implements the missing `data-rvt-disclosure-open-on-init` attribute for the disclosure component.

This attribute was listed in the documentation but never implemented.


levimcg (approved):
Haha, nice. Good catch 😂

PR #613, opened 2022-07-25
Renamed Modal component to Dialog

scottanthonymurray:
This PR renames the `Modal` component to `Dialog`.

The Cypress tests have been updated to reflect this change. The last Cypress test fails, since it's related to the functionality of keeping a dialog open when the user clicks outside of it. This functionality will be addressed in a separate PR.


levimcg (approved):
Looks good to me. I pulled it down and everything still works as expected 👍🏻

PR #614, opened 2022-07-26
Added positional CSS classes to dialog component

scottanthonymurray:
This PR adds four new CSS classes that can be used to modify a dialog component's position:

- `.rvt-dialog--top-left`
- `.rvt-dialog--top-right`
- `.rvt-dialog--bottom-left`
- `.rvt-dialog--bottom-right`

I experimented with other positional classes but they looked awkward and, seeing it, it was hard to imagine a use case for something like `.rvt-dialog--middle-right`.

### DIFF
--- a/src/sass/dialog/_base.scss
+++ b/src/sass/dialog/_base.scss
@@ -27,15 +27,39 @@
   padding-bottom: 5rem;
 
   &__inner {
-    position: relative;
-    top: 5rem;
+    position: absolute;
     max-width: 600px;
     width: 90%;
     background-color: $color-white-base;
-    margin: 0 auto;
-    margin-bottom: 5rem;
     border-radius: $spacing-xxs;
     overflow: hidden;
+    left: 0;
+    right: 0;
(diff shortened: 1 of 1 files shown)


scottanthonymurray:
@levimcg This CSS will be adjusted somewhat when I get to implementing the non-modal functionality so that the classes will only target that variant.

levimcg:
> This CSS will be adjusted somewhat when I get to implementing the non-modal functionality so that the classes will only target that variant.

@scottanthonymurray Cool, just so I'm 100% clear. Do you mean that the modifier class will target the root `.rvt-dialog` element for the non-modal version of the dialog? That's what I'm imagining in my head since for the non-modal version of the dialog (only) we'll need to undo the backdrop styling that is applied to the root element so the positioning would need to apply to the root `.rvt-dialog` element instead the `.rvt-dialog > .rvt-dialog__inner` element. Is that what you're thinking too?

scottanthonymurray:
Replacing these with position attributes.

PR #615, opened 2022-07-27
Implemented non-modal dialog as default behavior

scottanthonymurray:


levimcg:
@scottanthonymurray Just took a quick run through the changes locally. So far the only big thing I'm noticing is that in the modal dialog version (darkened background) if you try to click inside of the modal dialog when it is open, it closes it even when you click inside of the dialog element itself. The modal dialog should only close when someone clicks _outside_ of the dialog element, eg on the darkened background/document.

If i'm not mistaken, it looks like you'll need to check that if the click came from inside the dialog element in the `_clickOriginatedInsideDialogOrTrigger` method. right now you're checking if it came from inside using the `innerDialogSelector` property as the argument for `Element.closest()`, which I believe we removed, correct?

scottanthonymurray:
@levimcg Oh good catch! That should be fixed now.

levimcg (approved):
Great work on this, @scottanthonymurray! I can get started on implementing the remaining a11y fixes and then we can not talk about dialogs for a while, hopefully 😄.

PR #616, opened 2022-07-29
Implemented data-rvt-dialog-darken-page attribute

scottanthonymurray:


src/sass/dialog/_base.scss
```diff
@@ -30,7 +30,7 @@
   border-radius: $spacing-xxs;
   z-index: map.get($z-index, '1000');
 
-  &[data-rvt-dialog-modal] {
+  &[data-rvt-dialog-darken-page] {
     box-shadow: 0 0 0 9999px rgba(0, 0, 0, .85);
```
levimcg, line 34: Can we make this line use the rivet base black instead of plain 0, 0, 0 black using the sass `rgba()` function?

This line would become:

```sass
box-shadow: 0 0 0 9999px rgba($color-black-500, .85);
```

levimcg (approved):
One quick change to the color, but otherwise good.

PR #619, opened 2022-08-01
Added support to dialog component for multiple trigger buttons

scottanthonymurray:


src/js/components/dialog.js
```diff
@@ -558,7 +580,9 @@ export default class Dialog extends Component {
           return
         }
 
-        this.triggerButton.focus()
+        this.lastClickedTriggerButton
```
levimcg, line 583: This is kind of a weird edge case, but I feel like I remember running into a bug in the past where someone had application code that opened a modal with a trigger, and then by the time the modal was closed the trigger that initiated the opening had been removed from the DOM. I feel like that is a bit more likely to happen since we allow more than one trigger element.

It's a bit of a quirk, but `Element.querySelector` and `Element.querySelectorAll` both return a _static_ nodeList not a live one. That is the nodeList doesn't change even if the nodes/elements are moved from the DOM after the initial query takes place. So in this case `this.lastClickedTriggerButton` will always return true even if the trigger is removed from the DOM after opening and before closing.

Right now if you do that it just returns focus to the `<body>` element and doesn't throw, which is a decent fallback, but if we add an additional check to this ternary to make sure that the document.body still contains the trigger that opened it, this will work as intended and return focus to the first trigger element from the original nodeList as a fallback.

So the condition here would become:

```js
this.lastClickedTriggerButton && document.body.contains(this.lastClickedTriggerButton)
  ? this.lastClickedTriggerButton.focus()
  : this.triggerButtons[0].focus()
```

src/js/components/dialog.js
```diff
@@ -82,8 +82,10 @@ export default class Dialog extends Component {
       _initElements () {
         const dialogId = this.element.getAttribute(this.dialogAttribute)
 
-        this.triggerButton = document.querySelector(`[${this.triggerAttribute} = "${dialogId}"]`)
+        this.triggerButtons = document.querySelectorAll(`[${this.triggerAttribute} = "${dialogId}"]`)
```
levimcg, line 85: Can we wrap this `document.querySelectorAll(...)` in an `Array.from()`? Babel won't transpile a `.forEach` loop on a NodeList because forEach is implemented in all browser on Arrays, but not necessarily implement on the NodeList. It would be a bit safer here since we're using .forEach on this.triggerButtons later on to add the event listeners.

```js
this.triggerButtons = Array.from(
  document.querySelectorAll(`[${this.triggerAttribute} = "${dialogId}"]`)
)
```

levimcg (changes requested):
Couple of quick additions here. Let me know if you want to discuss.

PR #621, opened 2022-08-03
Added 2.0.0 production release workflow

scottanthonymurray:
This PR adds a GitHub workflow that triggers a production release when a branch is merged into `2.0.0-master`. A production release can also be forced manually via the `Actions` tab in GitHub.


levimcg (approved):
This looks good to me. I think the only thing we'll need to be careful of is automatically merging the automated _Dependabot_/security pull requests that get opened periodically.

Those get automatically opened agains the default branch which is currently the `2.0.0-master` branch. If we merge those it will kick off the release process. If that were to happen, we most likely would not have changed the package.json version number. The npm publish workflow will fail as a result of trying to publish a package with the same version number. But something to think about for the future.

PR #627, opened 2022-08-23
Improved accessibility of tabs and dropdown components

scottanthonymurray:


levimcg (approved):
@scottanthonymurray I updated node-sass to the latest version as discussed. I also removed the old build-test workflow since we aren't using it anymore.

PR #633, opened 2022-09-01
Updated dialog _clickOriginatedInsideDialogOrTrigger method to use event.composedPath().some()

### LINKED ISSUE #632: Event target being removed as a result of the click event causes dialog to close
## Describe the bug

Say you have a list of items and each item has some text and a "remove" button that when clicked removes that list item from the list, including the click event target (button) itself.

When the click event target itself is removed the logic that checks for clicks inside of the dialog is not longer satisfied because the event listener doesn't catch the click until the AFTER the event target is gone/removed from the DOM. This causes the dialog to close because as far as it's concerned, the click came from outside of the dialog element.

## Expected behavior

Dialog should remain open no matter what happens to the click event target unless explicitly closed programmatically by the developer.

- Reduced test case: https://codepen.io/levimcg/pen/BaxBMrZ
- There is a not in the pen (in the JS comments) about a workaround. It is possible to work around the issue in your own application logic by wrapping any calls to remove the click event target in a `setTimeout`, but this will most likely need to be fixed in the dialog's click event listener logic.

## Steps to reproduce

1. Click button to open dialog
1. Click any of the "Remove" buttons in the dialog body
1. Dialog closes, but should stay open after list item is removed from the DOM

scottanthonymurray:
Fixes #632

### DIFF
--- a/src/js/components/dialog.js
+++ b/src/js/components/dialog.js
@@ -333,8 +333,8 @@ export default class Dialog extends Component {
 
       _clickOriginatedInsideDialogOrTrigger (event) {
 
-        // This method uses event.path.some() to walk up the DOM tree to
-        // determine if the event target was within the dialog instead of
+        // This method uses event.composedPath().some() to walk up the DOM tree
+        // to determine if the event target was within the dialog instead of
         // this.element.contains(). It does so to prevent this method from
         // returning a false negative if a click event originating inside
         // the dialog removed its target from the DOM.
--- a/src/js/components/dialog.js
+++ b/src/js/components/dialog.js
@@ -332,8 +332,15 @@ export default class Dialog extends Component {
        ***********************************************************************/
 
       _clickOriginatedInsideDialogOrTrigger (event) {
-        return this.element.contains(event.target) ||
-               event.target.closest(this.triggerSelector)
+
+        // This method uses event.path.some() to walk up the DOM tree to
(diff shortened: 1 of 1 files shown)


src/js/components/dialog.js
```diff
@@ -332,8 +332,15 @@ export default class Dialog extends Component {
...
-               event.target.closest(this.triggerSelector)
+
+        // This method uses event.path.some() to walk up the DOM tree to
+        // determine if the event target was within the dialog instead of
+        // this.element.contains(). It does so to prevent this method from
+        // returning a false negative if a click event originating inside
+        // the dialog removed its target from the DOM.
+
+        return event.target.closest(this.triggerSelector) ||
+               event.path.some(el => el.dataset && 'rvtDialog' in el.dataset)
```
levimcg, line 343: This looks good and fixes the bug given the same test case attached to the original issue.

When I test it locally I get a console warning that `Event.path` is deprecated and will be removed with guidance to move to the replacement `Event.composedPath()`.

image


I tested this locally and `event.composedPath().some(...)` worked as expected. Let's go ahead and update this to use the new composedPath method instead.

levimcg (changes requested):
@scottanthonymurray One small update here, but otherwise this fix seems to be working great 👍🏻

levimcg (approved):
Great, thanks!

PR #639, opened 2022-09-26
Added repo filter to build-changelog workflow action

scottanthonymurray:
This PR fixes a bug with the `build-changelog` workflow action that caused it to fetch pull requests from *all* repos, rather than just `rivet-source`.


levimcg (approved):
Oh, nice! Thanks for looking into this 🙌🏻

PR #646, opened 2022-10-05
Removed Gulp

scottanthonymurray:
This PR removes Gulp as a dependency of Rivet and replaces all build tasks with plain Node scripts.

The `npm run start` task has been temporarily removed, as it will be replaced as part of the Eleventy branch.

(Note that this PR is being merged into `feature/env-refactor`.)


tasks/compile-js.js
```diff
@@ -0,0 +1,87 @@
...
+
+    /**************************************************************************
+     * Bundle, minify, and add license text to JS files.
+     *************************************************************************/
+
+    const bundle = await rollup.rollup({
+      input: './src/js/index.js',
+      plugins: [
+        nodeResolve(),
+        terser({
```
levimcg: Using the plugin here in the initial config makes it so that all bundles are minified once built which isn't the end result we want.

Is it possible to pass the terser plugin on a per bundle instance below. E.g. can that options object in `bundle.write({ ... })` take a `plugins:` key? I _think_ it can. That way you can just pass in terser on the bundle(s) that we do want to minify.

levimcg (changes requested):
This is looking great so far. See my inline comment re: minifying JS file. See table below for formats and what needs minified.

## Production design tokens

It looks like you haven't re-implemented the rest of the design tokens tasks that build them for production/distribution in the npm package. In the old gulp set up the compileTokens task would also build a `./tokens` folder that gets including the in the npm package. See unpkg: https://unpkg.com/browse/rivet-core@2.1.0/

That folder isn't getting generated. I would imagine that has something to do with when/where you are setting the `process.env.NODE_ENV` variable. I'm not sure?

## Production JS bundles

| Forma/file    | Expanded | Minified |
| ------------- |:--------:|:--------:|
| rivet-esm.js  |    ✅    |    🚫    |
| rivet-iife.js |    ✅    |    🚫    |
| rivet-umd.js  |    ✅    |    🚫    |
| rivet-min.js  |    🚫    |    ✅    |

levimcg (approved):
Nice, everything seems good to me so far 💯

PR #648, opened 2022-10-12
Updated components to refresh internal lists of child elements on DOM change

### LINKED ISSUE #647: Removing (programmatically) a Tab from DOM breaks keyboard (left/right) navigation of Tabs.
## Bug
When programmatically removing (and presumably adding) a tab (button and panel) from a rvt_tab, left/right keyboard navigation breaks. This manifests as not being able to keyboard navigate past the position from where the tab button was removed. 

## Expected behavior
Ideally keyboard navigation would continue to work when programmatically adding or removing a tab button and panel from the DOM. 

Internally this is caused by the way _initElements() initiates this.tabs. It creates a new shallow copy (using Array.from) of the tab list at init time, which is then used for navigation references, detached from what is in the actual DOM during usage. In turn if the DOM is programmatically changed by user actions to add or remove tabs, those changes aren't reflected in navigation. However mouse clicks function as expected due to not relying on this internal copy of the tabs (see onClick and _getClickedTabId). 

## Steps to reproduce
document.getElementById(tab-button-id).remove()
document.getElementById(tab-panel-id).remove()
Then attempt to navigate via keyboard.

scottanthonymurray:
Many components track an internal list of child elements, such as tabs or dropdown menu items.

However, this list was only built on first initialization, so if a child element (such as a tab) was added or removed, the internal list would fall out of sync and cause errors.

This PR updates several components to refresh their internal lists of child elements when the component's DOM changes. This is accomplished via `MutationObserver`.

Fixes #647

### DIFF
--- a/src/js/components/accordion.js
+++ b/src/js/components/accordion.js
@@ -148,23 +148,7 @@ export default class Accordion extends Component {
 
       connected () {
         Component.dispatchComponentAddedEvent(this.element)
-
-        this._watchForDOMChanges()
-      },
-
-      /************************************************************************
-       * Watches the component's DOM and updates references to child elements
-       * if the DOM changes.
-       *
-       * @private
-       ***********************************************************************/
-
-      _watchForDOMChanges () {
-        this.observer = new MutationObserver((mutationList, observer) => {
-          this._initElements()
-        })
-
-        this.observer.observe(this.element, { childList: true, subtree: true })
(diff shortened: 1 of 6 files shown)


levimcg (commented):
Nice! this is a great update! I played around deleting things from the DOM and everything seems to work great. 💯

levimcg (approved):

PR #662, opened 2022-11-16
Reduced width of current nav item red border to more closely match width of link text

### LINKED ISSUE #651: CSS issues in header
Hello I am a app developer for IU OVPUE. I have noticed in the rivet 2 header there is two aesthetic issues at my boss wanted me to report.

The first issue:
Underline extends to the right
Here the underline on the currently selected tab goes to far to the right.

The second issue:
No space at bottom of drop down
Here there is no padding at the end of the dropdown. I noticed that this is not the case until the screen width is greater than 740.

These are not urgent fixes.

Thanks,
Connor Hay

scottanthonymurray:
Fixes #651.

The problem in the linked issue with the padding under the identity module on mobile is not a Rivet CSS issue, but instead an error in the layout code. That will be fixed in a separate PR on the documentation site repo.

### DIFF
--- a/src/sass/header-system/_header-menu.scss
+++ b/src/sass/header-system/_header-menu.scss
@@ -196,6 +196,10 @@
       position: relative;
     }
 
+    &__item--current {
+      padding-left: 0;
+    }
+
     &__item--current &__link {
       color: $color-crimson-500;
     }
@@ -208,7 +212,7 @@
       content: '';
       display: block;
       background-color: $color-crimson-500;
-      width: 92.5%;
+      width: 100%;
       height: $spacing-xxs * 1.2;
       position: absolute;
       bottom: -$spacing-md * 1.1;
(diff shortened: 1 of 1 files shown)


src/sass/header-system/_header-menu.scss
```diff
@@ -208,7 +208,7 @@
       content: '';
       display: block;
       background-color: $color-crimson-500;
-      width: 100%;
+      width: 92.5%;
```
levimcg: This feels sort of like a magic number situation to me. I think this is actually being caused by some left padding that is applied to the nav item current state (`.rvt-header-menu__item--current`)

It looks like left padding is to account for the way the current state is displayed on small screens (red bar on the left of the nav item) and was accidentally left on at larger screen size too.

If we add a rule that removes the padding when the header nav switches from mobile to desktop layout, that will fix this issue where the current state on desktop looks wider than the nav item. It's being thrown off by that padding.

```scss
// This media query block already exists
@include mq($breakpoint-lg) {
  .#{$prefix}-header-menu {
    // existing nested styles
    &__item--current {
      padding-left: 0;
    }
  }
}
```

levimcg (changes requested):

levimcg (approved):
Looks good, thanks!

PR #667, opened 2022-12-21
Added new-component utility script for internal use

scottanthonymurray:
This PR adds a new internal utility script `new-component`. This script creates a boilerplate JS file for a new Rivet component.

The script takes the new component name, capitalized, as an argument:

```
npm run new-component Toast
```

Wrap component names with multiple words with double quotes to avoid ZSH syntax errors:

```
npm run new-component "Progress Bar"
```

This script uses the VS Code `code` command to immediately open the newly generated JS file in a new tab. For this to function, make sure that the `code` command is on your `PATH`. You can set it by going to **View > Command Palette**, typing `shell command`, and selecting "Install 'code' command in PATH".

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "move-example": "node ./tasks/move-example.js",
     "lint": "standard 'src/js'",
     "asset:upload": "npm run build && zip -r rivet-gh-release.zip css js sass tokens index.html",
+    "new-component": "node ./tasks/source-new-component.js",
     "sandbox-new-component": "node ./tasks/sandbox-new-component.js"
   },
   "files": [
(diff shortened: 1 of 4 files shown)


levimcg (approved):
Nice that worked like a charm. To expand/improve on this it would be nice if it also created a new folder inside of `./src/sass/[ component name ]` with a `_base.scss` and `_index.scss` file in it. Generally speaking, we create more component that only need CSS/Sass, so it would be super helpful.

scottanthonymurray:
@levimcg Updated the `new-component` script to also generate Sass files.

You can optionally pass `sass-only` to the end of the script if the component doesn't need JS:

`npm run new-component Callout sass-only`

PR #668, opened 2023-01-06
Refactored build process and development environment

scottanthonymurray:
This PR:

- Replaces Gulp with plain Node scripts
- Replaces Fractal with a simple Eleventy-based development sandbox


levimcg (changes requested):
Ok, this is a big one, but I think the only thing I can see that's left to finish up is a couple of small repetitive/unnecessary things in the CSS and JS development tasks and BrowserSync auto reload. We can touch base face-to-face if you want to work through these. I think after this you should be good to fix up any conflicts and then get it merged in! See below:

- You can update the sass watch task to just compile straight into `./dist/css/*` instead of building in the root `./css` directory every time. Then update the browserSync config `files: []` array to just look at the `./dist/css/rivet.css` file. that will get auto-reload for CSS working
- You can do the same thing for JS development task, but may need to create a separate `js-dev` (or whatever) task that just compiles the iife into `./dist/js/*`

Files array in Browsersync config in `.eleventy.js` should look like this in the end. Auto-reload should work after that.

```js
files: [
	'./dist/css/rivet.css',
	'./dist/js/rivet-iife.js'
]
```

levimcg (approved):
🚀

PR #672, opened 2023-01-24
Added switch component

scottanthonymurray:
This PR adds a Rivet 2-native implementation of Rivet 1's switch add-on.

### DIFF
--- a/src/sass/switch/_base.scss
+++ b/src/sass/switch/_base.scss
@@ -37,7 +37,6 @@
   &:focus {
     box-shadow: 0 0 0 math.div($spacing-xxs, 2) $color-black-000, 0 0 0 $spacing-xxs $color-black-400;
     outline: none;
-    transition: box-shadow .2s ease;
   }
 
   &[aria-checked="true"] {
--- a/src/sass/switch/_base.scss
+++ b/src/sass/switch/_base.scss
@@ -39,6 +39,10 @@
     outline: none;
   }
 
+  &:hover {
+    cursor: pointer;
+  }
+
   &[aria-checked="true"] {
     background-color: $color-green-500;
     border-color: $color-green-500;
--- a/src/sass/switch/_base.scss
+++ b/src/sass/switch/_base.scss
@@ -22,7 +22,7 @@
     background-color: $color-black-000;
     border-radius: $border-radius-circle;
     background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg' version='1.1'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M3.41421 2L7.91421 6.5L12.4142 2L13.8284 3.41421L9.32843 7.91421L13.8284 12.4142L12.4142 13.8284L7.91421 9.32843L3.41421 13.8284L2 12.4142L6.5 7.91421L2 3.41421L3.41421 2Z' fill='%234C5A69'/%3E%3C/svg%3E");
-    background-size: 16px 16px;
+    background-size: $spacing-sm $spacing-sm;
     background-repeat: no-repeat;
     background-position: center center;
     content: "";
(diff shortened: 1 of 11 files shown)


src/sass/switch/_base.scss
```diff
@@ -0,0 +1,128 @@
...
+  height: $spacing-xl + $spacing-xxs;
+  line-height: 1.1;
+  padding: $spacing-xxs;
+  position: relative;
+  -webkit-appearance: none;
+
+  &::after {
+    background-color: $color-black-000;
+    border-radius: $border-radius-circle;
+    background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg' version='1.1'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M3.41421 2L7.91421 6.5L12.4142 2L13.8284 3.41421L9.32843 7.91421L13.8284 12.4142L12.4142 13.8284L7.91421 9.32843L3.41421 13.8284L2 12.4142L6.5 7.91421L2 3.41421L3.41421 2Z' fill='%234C5A69'/%3E%3C/svg%3E");
```
levimcg: Let's ave Jenn re-export clean SVGs to use for the check and and "X" icons for the on and off state. She'll know what you're talking about, but I'll also give her a heads up in our 1:1 this morning. She just went through this process for the icon set so she should have the new production icons handy. We've got a process in place for exporting from Figma so that we don't end up with those `fill-rule` and `clip-rule` attributes in the SVG which can cause wonky display errors in certain stiuations.

src/sass/switch/_base.scss
```diff
@@ -0,0 +1,128 @@
...
+    height: $spacing-lg;
+    left: $spacing-xxs;
+    position: absolute;
+    top: $spacing-xxs;
+    transition: transform .2s ease;
+    width: $spacing-lg;
+  }
+
+  &:focus {
+    box-shadow: 0 0 0 $spacing-xxs/2 $color-black-000, 0 0 0 $spacing-xxs $color-black-400;
```
levimcg: I'm seeing Sass deprecation warnings for slash division syntax. Let's replace all instances in Sass file where `/` is used for math/division (it is deprecated). Update to new `math.div($spacing-xxs, 2)` or the native CSS `calc($spacing-xxs / 2)` function.

src/sass/switch/_base.scss
```diff
@@ -0,0 +1,128 @@
...
+    outline: none;
+    transition: box-shadow .2s ease;
```
levimcg: Let's remove the `transition` on the focus state (line 39). We did this with all the other form inputs/interactive elements recently, so let'd make this match those as well. The transition toggling from on to off should remain.

src/sass/switch/_base.scss
```diff
@@ -0,0 +1,128 @@
...
+  }
+
```
levimcg, line 45: I think in v1 the pointer cursor inherited from a rule we had set on the `button` element higher up in the cascade so now the hover state on the switch in v2 is just the default cursor. Let's add a hover state here after the focus state and use `cursor: pointer;` on hover like we did in the v1 version. You can see the original hover behaviour here for reference: https://indiana-university.github.io/rivet-switch/

src/sass/switch/_base.scss
```diff
@@ -0,0 +1,128 @@
...
+  line-height: 1.1;
+  padding: $spacing-xxs;
+  position: relative;
+  -webkit-appearance: none;
+
+  &::after {
+    background-color: $color-black-000;
+    border-radius: $border-radius-circle;
+    background-image: url("data:image/svg+xml,%3Csvg width='16' height='16' viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg' version='1.1'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M3.41421 2L7.91421 6.5L12.4142 2L13.8284 3.41421L9.32843 7.91421L13.8284 12.4142L12.4142 13.8284L7.91421 9.32843L3.41421 13.8284L2 12.4142L6.5 7.91421L2 3.41421L3.41421 2Z' fill='%234C5A69'/%3E%3C/svg%3E");
+    background-size: 16px 16px;
```
levimcg: Let's use design tokens here instead of hard-coded pixel values—should be `$spacing-sm`

levimcg (changes requested):
This is looking great. Some minor CSS/Sass updates to make and then I think we'll be good.

RE: getting it into the 2.3.0 release. Let's talk about this in our 1:1 today.

levimcg (approved):
Thanks for making these updates! 🚀

PR #674, opened 2023-01-25
Fixed bug causing dropdown component to open if link within dropdown element is clicked

scottanthonymurray:
The click event handler has been updated to check if the click event originated inside an `a` element inside the dropdown element. If so, the open/close state is not toggled.


levimcg (approved):
Nice, thank you 🙌🏻

PR #684, opened 2023-02-20
Added sandbox deployment scripts

scottanthonymurray:
The Rivet development sandbox can now be deployed to a private Sitehost account using the following commands:

- `npm run sandbox-deploy:test`
- `npm run sandbox-deploy:live`

An `.env` file for these deployment scripts has been provided via Secure Share.


levimcg (approved):
Awesome, this looks great

PR #686, opened 2023-03-02
Updated switch component to set aria-hidden="true" for on/off text labels

scottanthonymurray:


levimcg (approved):
@scottanthonymurray Nice, thank you! After you merge this would you mind deploying the sandbox so I can share for review? thanks!

PR #688, opened 2023-03-06
Updated Tabs component to auto-assign attribute values if none are provided

scottanthonymurray:


src/js/components/tabs.js
```diff
@@ -359,6 +490,205 @@ export default class Tabs extends Component {
...
+          { tab, panel }
+        )
+
+        return dispatched
+      },
+
+      /************************************************************************
+       * Removes a tab with the given ID or index value.
+       *
+       * @param {string} idOrIndex - ID or index of tab to remove
```
levimcg: Just so we match the the `activateTab` method, should we update the JSDoc param here to `@param {string|number}`?

levimcg (approved):
One tiny comment here re: JSDoc comments on the `removeTab` method. Otherwise this looks great to me!

PR #693, opened 2023-04-19
Updated component icon data URIs

scottanthonymurray:
This PR updates component icon data URIs in `background-image` properties. Many component icons are actual `svg` elements, which will be updated in the documentation site examples.


levimcg (changes requested):
Nice, this looks good. The only thing missing is the Disclosure component. It uses a right-facing chevron icon as a background image in the `::before` pseudo element. Once you get that updated with the new icon, this should be good to go.

levimcg (approved):

PR #694, opened 2023-05-02
Set colors of rvt-sr-only elements to prevent a11y tool false positives

scottanthonymurray:


levimcg (approved):
Nice, that seems to have fixed the false positive on the footer when I test it with the WAVE browser extension 👍🏻

PR #698, opened 2023-05-22
Renamed generated zip file in asset:upload npm script

scottanthonymurray:
Renamed `rivet-gh-release.zip` to just `rivet.zip` so that we don't have to rename it when uploading to the releases page.


levimcg (approved):
💯

PR #704, opened 2023-06-20
Updated accordion component to auto-assign trigger and panel IDs

scottanthonymurray:
Similar to tabs, this PR updates the accordion component to auto-assign trigger and panel ID attributes if none are provided. Also generates appropriate `aria-labelledby` attributes.


levimcg (approved):
💯

PR #728, opened 2023-10-30
Fix badge style

basham:
This PR fixes:

1. Height is now effectively `24px` (rather than `27.52px`).
2. Font size is now effectively `14px` (rather than `88%`).
3. Pixel units are replaced with `rem`.
4. Badge content should not wrap, because it should be succinct.
5. Secondary badges now have a white background, to maintain color contrast on non-white backgrounds.

Notes:

1. This removes the ability to scale badges according to its context. However, I've never seen such a use case. That can just lead to some odd uses. If there should be different sizes, there should be new modifiers.
2. Is it reasonable to ensure the badge does not wrap?
3. Values made with `calc()` are compiled by Sass, so it reduces runtime calculations. Neat.

### DIFF
--- a/src/sass/badges/_base.scss
+++ b/src/sass/badges/_base.scss
@@ -3,33 +3,25 @@
 
 @use '../core' as *;
 
+$border-radius: 999rem;
+
 .#{$prefix}-badge {
   background-color: $color-black-100;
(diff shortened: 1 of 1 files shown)


levimcg (approved):
This looks great!

PR #730, opened 2023-11-07
Switch: Adjust height, add medium variant

basham:
Notes:

1. Adjusted the default height from `44px` to `40px` to match the height of buttons.
2. Added a `medium` variant at `32px` in height.
3. Cleaned up styles to minimize overrides for variants.
4. Replaced "On" icon transition from `translate` to `left`. This allows it to animate flush right regardless of the width of the button container. With `left`, percentage values are relative to its container. With `translate`, percentage values are relative to itself.
5. This has been visually tested in Chrome, Firefox, and Safari.

### DIFF
--- a/src/sandbox/components/switch/variants/config-page.njk
+++ b/src/sandbox/components/switch/variants/config-page.njk
@@ -4,7 +4,7 @@ title: Switch configuration page
 layout: layouts/preview.njk
 permalink: /components/preview/{{ title | slug}}/
 padding: true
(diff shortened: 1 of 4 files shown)


basham:
- Comment out medium size variant, and save it for later. This may be part of a larger rework of buttons and links.

basham:
@levimcg: Ready for review.

levimcg (approved):
Great, thanks!

PR #732, opened 2023-11-09
Button style enhancements

basham:
Notes:
1. Added white backgrounds to secondary buttons, so contrast is maintained on non-white backgrounds. See similar changes in #728.
2. Button content should not wrap, because it should be succinct. See similar changes in #728.
3. Reduced left and right button padding by `2px` on each side, so that normal-sized icon buttons are 40 square pixels.
4. Added `rvt-button--medium` variant, with `32px` height.
5. Added new medium and small variant examples in the "Button with icons" page.

Questions:
1. Should buttons use gap for icon button spacing, instead of `rvt-m-left-xs`? Or does that presume too much about the contents of buttons and how they should be used?
2. The `2px` horizontal padding for small buttons seems insufficient, in order to ensure that small icon buttons remain 24 square pixels. Does this indicate that perhaps the padding should be dynamic, according to if the visible content is an icon only verses icon with label verses just label?
3. We may want to consider later enhancing icon buttons so that there is a tooltip-style label that shows when it is in focus or on hover. The problem is how do we control the placement of such a thing? The new [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) would be great, but it is not yet available in Firefox. The upcoming [anchor positioning CSS property](https://developer.chrome.com/blog/introducing-popover-api/) is still being [drafted](https://drafts.csswg.org/css-anchor-position-1/). So, it would likely require the user to just pre-determine the position (bottom or top) and call it good enough.
4. Perhaps I should do more exemplar research. But the [Adobe Spectrum button](https://spectrum.adobe.com/page/button/) has a lot of useful variants.

### DIFF
--- a/src/sandbox/components/button/preview/button-full-width.njk
+++ b/src/sandbox/components/button/preview/button-full-width.njk
@@ -4,7 +4,7 @@ title: Full-width button
 layout: layouts/preview.njk
 permalink: /components/preview/{{ title | slug}}/
 padding: true
-order: 4
+order: 5
 ---
 <button class="rvt-button rvt-button--full-width">Primary</button>
 <button class="rvt-button rvt-button--success rvt-button--full-width">Success</button>
--- a/src/sandbox/components/button/preview/button-group-right.njk
+++ b/src/sandbox/components/button/preview/button-group-right.njk
@@ -4,7 +4,7 @@ title: Right-aligned button group
 layout: layouts/preview.njk
 permalink: /components/preview/{{ title | slug}}/
 padding: true
-order: 6
+order: 7
 ---
 <div class="rvt-button-group rvt-button-group--right">
   <button class="rvt-button">Primary</button>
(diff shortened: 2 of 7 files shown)


basham:
I've made this is PR a draft until the questions are addressed. We may need a meeting to discuss the direction.

levimcg:
@basham Cool! Thanks for getting this open. Yes, I'd be happen to schedule something face-to-face to chat through some of this before we continue.

basham:
- Comment out medium size variant, and save it for later. This may be part of a larger rework of buttons and links.
- Add a gap of `xs`. This will provide default spacing for icons and the text content, so we don't have to use the margin utility.
- Adjust the width of small buttons so that the icon button is `32px` wide. This provides a more comfortable look, even though it means they are not square (like the large and medium buttons).

basham:
@levimcg: Ready for review.

levimcg (approved):

PR #743, opened 2024-02-19
Integrate CSS custom properties into core

basham:
Notes:

1. Should the generated file base name be `css-custom-properties`?
2. Should the generated file extension be `.scss` or `.css`? Either works, since it doesn't produce any Sass-specific syntax.
3. Should the platform name in the config be `src/sass/core/css`? The [platform name](https://amzn.github.io/style-dictionary/#/config?id=platform) can be anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ cypress/screenshots
 /src/sass/core/maps/*.scss
 !/src/sass/core/maps/_index.scss
 /src/sass/core/variables.scss
+/src/sass/core/css-custom-properties.scss
--- a/.tokens.config.js
+++ b/.tokens.config.js
@@ -13,6 +13,17 @@ module.exports = {
         }
       ]
     },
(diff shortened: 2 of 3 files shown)


levimcg:
@basham This is perfect. I'll share my answers to your questions here.

1. I'm fine with calling the file that. I think that `custom-properties.css` would be fine too, but the `css-` at the beginning makes it ultra explicit what the file contains.
2. I'd say let's keep it `.scss` just since we don't use the `.css` extension anywhere else in the `.src/sass` directory or the `./sass` directory that gets shipped with the npm package.
3. I think that name is fine. I guess it could also be `src/sass/core/css-custom-properties` for consistency, if dashes are allowed.

basham:
@levimcg I updated the platform name as you suggested. Dashes and any string value is allowed. So, this is ready to merge, whenever you're ready.

PR #761, opened 2024-07-01
Updated dialog to ignore hidden elements when determining focusable children

### LINKED ISSUE #760: Dialog fails to trap focus when it contains focusable elements that are temporarily hidden
## Describe the bug
If a dialog contains any focusable elements that are temporarily hidden, it still uses the hidden elements when calculating all of the focusable elements inside of itself. This means that in some cases it cannot properly trap focus and a user can use the Tab key to navigate out of the dialog while it is open.

A good example of this would be if a dialog contains a dropdown menu or a disclosure of some sort that has focusable elements inside. If say, the last focusable element is hidden inside a dropdown menu, the dialog doesn't know that it's the last element which means it doesn't know to move focus back to the first focusable element allowing you to navigate out of the focus trap.

## Expected behavior
The dialog should only use visible focusable element when determining how to trap focus.

- Link to a test case or any CSS, JS, HTML relevant to the bug.
- Add any relevant error messages thrown by the browser.
- If applicable, add screenshots to help explain your problem.
- The browsers in which you encountered the bug, if applicable.
- Add any other context about the problem here.

## Steps to reproduce
Here is a [reduced test case](https://codepen.io/levimcg/pen/Yzbbxpw) to replicated the bug.

1. Open the dialog
1. Use the Tab key to navigate through the dialog
1. After navigating past the dropdown toggle button, you will move out of the dialog instead of focus being trapped inside.

scottanthonymurray:
Fixes #760

### DIFF
--- a/src/js/components/dialog.js
+++ b/src/js/components/dialog.js
@@ -407,7 +407,7 @@ export default class Dialog extends Component {
         )
 
         this.focusableChildElements = Array.from(this.focusableChildElements)
-        this.focusableChildElements = this.focusableChildElements.filter(el => el.checkVisibility()) // Exclude hidden elements
+        this.focusableChildElements = this.focusableChildElements.filter(el => el.clientHeight > 0) // Exclude hidden elements
 
         this.firstFocusableChildElement = this.focusableChildElements[0]
         this.lastFocusableChildElement = this.focusableChildElements[this.focusableChildElements.length - 1]
--- a/src/js/components/dialog.js
+++ b/src/js/components/dialog.js
@@ -406,6 +406,9 @@ export default class Dialog extends Component {
           'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="-1"]'
         )
 
+        this.focusableChildElements = Array.from(this.focusableChildElements)
+        this.focusableChildElements = this.focusableChildElements.filter(el => el.checkVisibility()) // Exclude hidden elements
+
         this.firstFocusableChildElement = this.focusableChildElements[0]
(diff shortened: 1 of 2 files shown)


src/js/components/dialog.js
```diff
@@ -406,6 +406,9 @@ export default class Dialog extends Component {
           'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="-1"]'
         )
 
+        this.focusableChildElements = Array.from(this.focusableChildElements)
+        this.focusableChildElements = this.focusableChildElements.filter(el => el.checkVisibility()) // Exclude hidden elements
```
levimcg: This is one of those gray areas where we may need to look at browser support/adoption a bit more closely. the `Element.checkVisibility()` while it technically is available in all modern/evergreen browsers, it is _super_ new. My biggest concern here is that it was only release ind Safari (macOS and iOS) back in May 2024. Knowing anecdotally that a lot of iOS and macOS users tend to not always update right away because it take manual action to do so (unlike Chrome, Firefox, etc.), we may want to wait a bit before we start using this method in production. Especially since we have such high traffic on our sites from iOS Safari.

We'll want to go with something simpler like checking the clientHeight, etc.

levimcg (changes requested):
We can talk more about this face-to-face if you want just let me know.

src/js/components/dialog.js
levimcg: @scottanthonymurray Sorry, meant to share this link to the browser stats

https://caniuse.com/mdn-api_element_checkvisibility

levimcg (approved):
This looks good to go! 💯

PR #763, opened 2024-07-10
Updated dependencies for July 2024

scottanthonymurray:


package.json
```diff
@@ -74,6 +74,7 @@
     "style-dictionary": "^3.1.1"
   },
   "dependencies": {
+    "@babel/traverse": "^7.23.2",
```
levimcg: This is weird that this got added to the regular dependencies object? Anything babel related is just for transforming code and would never get shipped with the actual package. E.g., our other babel deps are in `devDependencies`. Did you have to manually install this to fix something?

scottanthonymurray: Oh weird, not sure why that happened. I didn't manually do that, `audit fix` must have done that on its own. I'll get that taken care of here before the `2.8.1` release.

basham:
@levimcg Can we just close this PR, since this is out of date? All the deps are updated in the `v3-main` branch. Is that good enough?

levimcg:
@basham Yep, sounds good to me. I'll close this now.

PR #765, opened 2024-07-30
Removed aria-haspopup attribute from sidenav and dropdown

scottanthonymurray:
This PR removes the `aria-haspopup` attribute from the sidenav and dropdown components, with the exception of dropdown components used as application menus.

### DIFF
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -81,6 +81,16 @@ export default class Dropdown extends Component {
       },
 
       /************************************************************************
+       * Initializes dropdown state properties.
+       *
+       * @private
+       ***********************************************************************/
+
+      _initProperties () {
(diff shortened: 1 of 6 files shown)


levimcg:
@scottanthonymurray Not sure if you're working on it, but could you also remove the `role="menu"` on this line in the example "application" menu example (with buttons). https://github.com/indiana-university/rivet-source/pull/765/commits/9edd732658ab90d2f28cc55bde83b5ad0ba6089c#diff-4e84df8e0f1495b01d8b57e65acb9a8a251b660b9869edafe8769331e103e3e0L16

similar to the aria-haspopup with the button, we should let the JS handle adding the attribute here in this example so we can test it properly.

levimcg (approved):

PR #767, opened 2024-11-21
Added first draft of Priority Nav component JavaScript

scottanthonymurray:
@levimcg First take of the priority nav component. Still needs refinement for sure--I like your note about having the "More" menu markup being generated by the component so that a developer doesn't have to copy paste in all that empty boilerplate for it.

Preview is in the sandbox under the Header component. I'll eventually move it out into its own sandbox item.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "rivet-core",
-  "version": "2.8.0",
+  "version": "2.8.1",
   "lockfileVersion": 2,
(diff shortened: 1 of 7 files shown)


levimcg (approved):
Sweet, this is a great start! I agree, let's explore generating the markup for the "more" button/menu on init. It only works with JS, so it makes sense to only have the markup there if JS is available. I think moving it into the actual header context is a great next step.

PR #770, opened 2024-11-26
Added support for data-rvt-dropdown-open-on-hover attribute

scottanthonymurray:


levimcg (approved):
@scottanthonymurray Nice this is great. The only minor thing I would say is to removed the mouseenter and mouseleave event handlers in the `_removeDocumentEventHandlers` method just so they get cleaned up of if someone is for some reason rendering the header client-side. Otherwise, works exactly as expected 👍🏻

PR #771, opened 2024-12-03
Implemented collapsible sidenav

scottanthonymurray:
Implemented the first pass of the collapsible sidenav.

You can view an example at:

`http://localhost:8080/layouts/preview/base-layout-two-column/`

Code needs refining but it should be functional!

### DIFF
--- a/src/js/components/component.js
+++ b/src/js/components/component.js
@@ -177,6 +177,25 @@ export default class Component {
   }
 
   /****************************************************************************
+   * Watches the document body for resize events. Accepts a callback to
+   * perform additional updates to the component on document body resize.
+   *
+   * @static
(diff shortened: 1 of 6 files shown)


levimcg (approved):
Nice, this is great for a first pass! Let's talk more about being able to configure the breakpoint via data attribute and whether that's something we want to implement.

scottanthonymurray:
@levimcg Thanks! This first version of the collapsible sidenav accepts an `em` value as an optional value for the data attribute, so if you did something like:

`data-rvt-sidenav-collapse-on-mobile="18em"`

It would use that breakpoint instead of the default `rvt-breakpoint-lg`.

It's super easy to snip this out if we ultimately decide we don't want to allow this functionality.

PR #778, opened 2025-03-03
Removed automatic application of ARIA attributes to dropdown component

### LINKED ISSUE #775: Do not require role="menu" on the Rivet Dropdown component
## Describe the bug
We use the Rivet Dropdown component in multiple tools for filtering which content is displayed for the user.  These dropdowns utilize checkboxes, radio buttons, and buttons (for clearing the filters).  We purposely do not include the role="menu" on these dropdowns.  While doing accessibility testing, we were advised that the role was unnecessary and actually made them less accessible. 

In a recent release of Rivet, this role is now being added via javascript, breaking accessibility on all of our dropdowns.  Adding the appropriate role="menuitemcheckbox" or role="menuitemradio" on the menu items allows it to pass an accessibility validator, but there are still issues with navigation and the user experience.  When you add role="menu", you are committing to that component behaving in a very specific way.  

There are several open issues related to navigating the menu that are also affecting us now that role="menu" is added to our dropdowns:
The dropdown component's tab and shift+tab handling violate ARIA practices for menuitems
https://github.com/indiana-university/rivet-source/issues/599
Handle the "optional" ARIA menu and menuitem keyboard interactions
https://github.com/indiana-university/rivet-source/issues/601

Until these are addressed, these dropdowns will not be fully accessible menus.

I will also add that if you follow the "Dropdown with Buttons" example in the Rivet documentation, the current UI doesn't work well for the menuitemcheckbox or menuitemradio roles.  This old discussion post recommending the addition of checkmarks (and I would add a radio indicator) still applies:
https://github.com/indiana-university/rivet-source/discussions/420
This example in the ARIA guide uses both:
https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/

## Expected behavior
Do not require role="menu" and aria-haspopup on the Dropdown component.  That decision needs to be made by the team implementing it, especially since they also need to include the appropriate roles for the menu items.  If the issues with menu navigation and UI are addressed, then the requirement can be re-evaluated.

See the open issues referenced above.

## Steps to reproduce
https://rivet.iu.edu/components/dropdown/?example=dropdown-with-buttons
In the "Dropdown with Buttons" example in the Rivet documentation, if you inspect the source code, you will see that role="menu" has been added to the div containing the menu and aria-haspopup="menu" was added to the "Dropdown with buttons" toggle button.  These are not added for the Dropdown examples with links.

scottanthonymurray:
Fixes #775

### DIFF
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -130,21 +130,6 @@ export default class Dropdown extends Component {
       },
 
       /************************************************************************
-       * Determines if the dropdown is a menu for the purposes of ARIA
-       * attributes.
-       *
-       * @private
-       * @returns {boolean} Dropdown is menu
-       ***********************************************************************/
-
-      _isMenu () {
-        // Consider the dropdown a menu of action buttons if at least one
-        // non-hyperlink element is present
-        
-        return this.menuItems.some(i => i.tagName != 'A')
-      },
-
-      /************************************************************************
        * Removes the arrow icon from the tab order.
        *
        * @private
--- a/src/js/components/dropdown.js
+++ b/src/js/components/dropdown.js
@@ -112,7 +112,6 @@ export default class Dropdown extends Component {
 
       _initAttributes () {
         this._assignComponentElementIds()
-        this._setAriaAttributes()
       },
 
       /************************************************************************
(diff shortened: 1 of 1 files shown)


levimcg (approved):
Approved once unused `_isMenu()` method is removed

PR #785, opened 2025-08-19
Add new color tokens

basham:
Pushing this small piece of the `v3-color` branch work before the rest is ready, so it can be used in other work this week.

### DIFF
--- a/src/tokens/colors.json
+++ b/src/tokens/colors.json
@@ -1,5 +1,119 @@
 {
   "color": {
+    "base": {
+      "dark": {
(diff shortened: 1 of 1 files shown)


levimcg (approved):
Nice, this looks like a good start to me!

Just curious, is the plan to leave all of the old number-based variants in place as you work through updating each component, etc.?

basham:
Yes, I'm leaving in all the old tokens for now until I update and clean up the references. I want to try to avoid breaking things as much as I can!

PR #786, opened 2025-08-26
Integrate new colors

basham:
Notes:
1. Added new colors as tokens.
2. A lot of lines changes were due to changing CSS rule order to alphabetic. I should have held this off for a different PR.
3. Replaced all Sass color variables with native CSS custom properties.
4. Component pages repeat the same HTML for each of the three color themes. This makes it easier to see how a particular component adapts for the themes, instead of seeing them one at a time.
5. Icons are now implemented with `clip-path: path("…");` rather than with `background-image`. This approach allows the icon color to be controlled by `background-color` and significantly saves space.
6. In a future PR, these icons should be abstracted as tokens, so they become CSS variables. That will save even more space.
7. Removed `rvt-accordion__toggle-icon`.
8. Removed depreciated `rvt-accordion__icon-bar`.
9. Removed `rvt-badge--secondary*` variants.
10. Added `rvt-tag` component (an interactive alternative to a badge, which needs more discussion).
11. Removed depreciated `rvt-breadcrumbs--call-out` variant.
12. Removed `rvt-seriesnav__icon`.
13. Removed `rvt-callout--crimson` in favor of using themes.
14. Removed `rvt-footer-base--light` in favor of using themes.
15. Removed `rvt-link--reverse` and `rvt-link-reverse` in favor of using themes.
16. Removed `rvt-border-color-*` classes.
17. **Breaking:** Changed `<select class="rvt-select">` to `<div class="rvt-select"><select>`.
18. I will remove the old color tokens, `rvt-color-*`, and `rvt-bg-*` classes in another PR. That will be big, and this PR is big enough.
19. Further refinement is needed on how the colors are applied, such as the use of validation messages and field styles.
20. Every component needs to be reviewed for quality assurance. But that could wait until a few more PRs for further clean up.


levimcg (commented):
I had some extra time yesterday and this morning so I was able to review this! 

This is really excellent work. I can tell you all have put a lot of thought and consideration into these updates and I think it's going to set us up with a great foundation to continue building on!

I have a couple of notes here, but fully realize this is the first part of a significant update that will take a few more PRs to continue to refine. Other than these few things I this looks ready to me. I would like to get this merged soon so that you all can move on to smaller more focused PRs on the remaining things you mentioned in the original post. Here are my notes:


## Brand/Crimson Buttons

**Background info**:
In the past we tried to draw a divide between button elements (`<button>`) and call-to-action links that look like buttons— the [Button component](https://rivet.iu.edu/components/button/) and the [Call to action component](https://rivet.iu.edu/components/call-to-action/). Technically this seems like the best way to distinguish semantic purpose of the button element from a link that looks like a button, but in practice it meant that people were looking for a branded CTA button in the Button component docs and rarely found the actual CTA component (link that looks like a button) they were actually looking for. This is why we originally decided to fold the CTA component styles into the Button component as the default variant in the lates v3 alpha. It's nearly universal that people think of what is actually a CTA *link* (`<a>` element) as *buttons* because it's standard practice to style them that way. Especially in the marketing world.

I'll leave the implementation up to you and Jenn to collaborate on, but **here are the requirements**. I would address this as a separate PR unless it makes sense to do it as a part of this one.

- We need a brand/crimson variant of the button that is crimson for link calls to action on marketing sites. All of these design have been approved in collaboration with the marketing team using Crimson CTA links (links that look like buttons)
- The brand/crimson button should have the same variants as the default button: default (with outline), solid, and plain only using the brand crimson for the color
- The dark and crimson themed versions can follow the same color patterns you've already defined for the default button, but that's up to you all.
- `:visited` pseudo class on buttons needs a bit more work when used in themes in composite components like the hero. This is an issue I'm seeing with the themed buttons in general.

## Billboard

- Dark and Crimson variants missing. The crimson variant will be important since it it is used extensively in Site Builder

manselldesign:
@levimcg You might be saying this but I'm wondering if the easiest way to distinguish the button colors here is the default (black button) is the **button** variant and the brand (crimson) variant is the **CTA button**? Basically all app elements use the black button and marketing elements use the crimson CTA button.

basham:
I fixed the billboard color for the various themes. I would like to hold off on the CTA changes for another PR.

levimcg:
@manselldesign Yeah, I think that could work. Ultimately, I'll leave implementation details to you all. I think making a clear distinction that CTA/Brand buttons are only for links and not necessarily for building forms, etc. is a good approach.

levimcg (approved):
Looks great! I'm approving so we can work through additional updates in smaller increments now that we have a good foundation in place.

PR #787, opened 2025-08-29
Badge updates

basham:
Notes:
1. Changes incorporate feedback from the UXD group and UX CoP from this last week.
2. Expect further discussion and updates to badges.
3. This change restores the idea of secondary badges that was removed in #786.
4. The lighter-colored "secondary" badges should be considered the "default" type compared to the darker-colored badges.
5. Badges are only static. They should never be interactive, as a link or button.
6. The solid background color and rounded corners of the badge provides contrast against the button aesthetic, which by default is outlined with sharp corners.
7. In order to enforce proper semantics, the light badges can only be used like `<span class="rvt-badge">` while the dark "stronger" badges can only be used like `<strong class="rvt-badge">`.
8. There is not enough distinction between a button and the tag concept. So, the tag component that was introduced in #786 is replaced with the button variant `rvt-button--tag`.
9. This work also adds a pressable button style with `aria-pressed="true/false"`, because this idea was explored in tandem with these badge/tag discussions.
10. Buttons will be further refined in a future PR.
11. The danger colors are updated to provide more contrast against the crimson/accent color.
12. The "x" icon used for button tags will also be updated in a future PR. For whatever reason Figma was not exporting this new icon right, and its code is excessively big.


levimcg (approved):
This looks great. I like the idea of using semantic elements to distinguish between the secondary and default badges, really smart!

PR #788, opened 2025-08-29
Upgrade dependencies

basham:
Note:
1. This required updating some major deps, so that required some more adjustments.
2. I changed all dependencies to use exact versions. This guarantees that everyone who runs the repo will be running the exact same thing.
3. I think the build environment now starts up a little faster.


levimcg (approved):
Looks good. Everything runs locally for me as expected. I did notice a couple of Sass deprecation warning, but otherwise the build and dev/serve process do seem a lot faster!

basham:
Yes, I can work on the depreciation warnings on a another PR.

PR #789, opened 2025-09-02
Fix Sass depreciation warnings

basham:
Follow up to #788.

To resolve the [mixed declarations warning](https://sass-lang.com/documentation/breaking-changes/mixed-decls/), I just needed to declare the `@include` rules after the other rules.


levimcg (approved):
Nice, warnings are gone now 👍🏻

PR #790, opened 2025-09-02
Update tokens to use DTCG format

basham:
Notes:
1. This work was prompted by the shadow tokens not generating as expected since the dependency upgrade (#788).
2. I updated the tokens to align with the new [Design Tokens Format](https://www.designtokens.org/tr/drafts/format/), which Style Dictionary mostly supports.
3. [Once supported](https://github.com/style-dictionary/style-dictionary/issues/1501) (ideally in an upcoming minor upgrade), the dimension tokens should be updated from `"$value": "1rem"` to `"$value": { "value": 1, "unit": "rem" }`.
4. I updated the token group "font" to "font-family" to align with the new convention.
5. The "shadow" type is considered a composite value, which allows the sub-properties to be explicitly declared (offsetX, offsetY, blur, spread, color) and make it more readable.
6. Future work: Replace all Sass variable instances with CSS custom properties.
7. Future work: Add icon and sticker tokens, as their raw path values. We can write transformations so that the corresponding SVGs are an output of the Style Dictionary instead of an input!


levimcg (approved):
Looks good 👍🏻

PR #791, opened 2025-09-03
Make icon tokens

basham:
Notes:
1. The icon token values must be wrapped in a second layer of quotes to make sure that the output is a string (`"$value": "'…'"`).
2. Future work: Develop a Style Dictionary transformation that will remove the need for the second layer of quotes and automatically wrap the value in `path()` — since that's the only way we use it in CSS.
3. Change: Removed `rvt-social-link--twitter` in favor of just `rvt-social-link--x`. We shouldn't keep old brand references, especially if we are not keeping the old logos.
4. New icons: `close-small`, `exclamation`, `info`, `logo-iu`, `quote-open`, `social-*`, `text`
5. Future work: Add all icons from the [Rivet Icons repo](https://github.com/indiana-university/rivet-icons).
6. Future work: Migrate all uses of inline SVG in components to using CSS icons.


levimcg (approved):
Nice work, this is very cool!

PR #793, opened 2025-09-04
Work on cards, tags, and colors

basham:
Notes:
1. Added `rvt-button--dismissable` which will display a small "x" after the button content.
2. Refined card content and spacing with image, sticker, and plain variants.
3. Adjusted base colors.


levimcg (changes requested):
This looks good. One issue I'm seeing with some spacing if certain elements are not present in the card body.

If the card has tags but does not have meta/date information, the spacing gets tight between the tags and the rest of the card body content. This could be mitigated with template logic in a real use case by conditionally adding a spacing utility class to the tag element, but it would be nice if it was automatically spaced no matter the combination of elements present.

Screenshot below.

image

basham:
I was much too quick to set this work as ready for review. There's a lot more work on the cards we need to do, including the work you mentioned. I will revert this PR back to draft.

basham:
This PR is finally ready for review again.

Notes:
1. Simplified the 11ty sandbox, requiring less metadata.
2. Added new serif Playfair Display.
3. Allow `.rvt-avatar` to be resized with CSS (as what happens in the Profile Card) — not just with HTML.
4. Added new `.rvt-sidenav--compact` class for the sake of the sandbox navigation.
5. Updated base color tokens.

Future work: 
1. Replace custom "sections" 11ty collection with 11ty Navigation plugin.
2. Apply new sandbox configuration to all sandbox files, not just the Card component files.
3. Create more abstractions in `macros/card` to reduce repeated content and allow more content variations.
4. Review the `.rvt-sidenav--compact` variant.
5. Review `.rvt-button--tag` and `.rvt-button--dismissable`.
6. Refine breakpoints, especially the Feature Card.

levimcg (approved):
This looks great! Let me know if/when you're ready to work another alpha release that we can test out in Site Builder. 🎉

PR #794, opened 2025-09-04
Use Vite for the dev environment

basham:
Notes:
1. With Vite, the dev environment now starts up within 2 seconds!
2. Use `npm start` to start the dev environment.
3. `src/sandbox/rivet.js` is the entry point for Vite to know what to watch and compile.
4. Future work: `npm run build` still uses the current Rollup tasks. This will be replaced with `vite build`. After that, many dev dependencies and custom Node tasks can be removed.


basham:
Just noticed that the JavaScript components are not working. Will work on a fix in this PR.

levimcg (approved):
Nice, the dev sandbox is much faster to start up. Looks like a good start to me. Seems like you have a plan for building CSS, JS and other production/npm package related stuff from the `./src/*` folder.

basham:
@levimcg Please review this again.

Notes:
1. The Rivet JS now loads in the sandbox environment.
2. Run `npm run build:vite` to see new `/dist` output files. These are meant to eventually replace the `/css` and `/js` output files. `npm run build` still builds the old files for now, so you can compare them.

levimcg (approved):
This all looks good as far as I can tell. It's cool that we can just use one JS entry point to compile everything including the sass files. JS components are working in the sandbox now.

The `dist` folder output looks good to me. I think we'll want to have a conversation about what the final dist/output looks like in terms of structure and what files we'll continue to include in the npm package, but it sounds like you have a plan there.

PR #795, opened 2025-09-05
Refine dist output

basham:
Notes:
1. Removed all old tasks and dependencies used to generate out files.
2. The new `dist` folder contains `rivet.css`, `rivet.js`, `rivet.js.map`, and `tokens.json`.
3. Needs further research: I think the `module`, `unpkg`, and `export` keys are done correctly in [`package.json`](https://docs.npmjs.com/cli/v11/configuring-npm/package-json), but we should somehow verify this configuration.
4. Needs further research: All dependencies are up to date, but `npm audit` complains about [`vite-plugin-eleventy`](https://github.com/Snugug/vite-plugin-eleventy), which hasn't been updated in four years. Because it just affects the dev environment, I'm not too concerned. But perhaps there is a more updated alternative to try.
5. Needs further research: The `rivet.js` file seems bigger than what it should be (153kb) but it at least gzips well (19kb). Vite purposefully doesn't aggressively [minify ESM files](https://vite.dev/config/build-options.html#build-minify) ("Note the build.minify option does not minify whitespaces when using the 'es' format in lib mode, as it removes pure annotations and breaks tree-shaking."). I tried configuring Terser as the suggested alternative, but I didn't see any affect, so I'm moving on for now.


levimcg (approved):
This looks good to me. Looks like there is an official 11ty/vite plugin. Not sure if that's an option here, but may be worth a look. https://www.11ty.dev/docs/server-vite/

I'm all for only shipping an ES module version of rivet.js, but if it's not possible to minify the whitespace we will need to figure out an approach where we can ship a truly minified version for folks that are not able to use their own build process to bundle/minify (basically anyone using the WCMS). Marketing sites tend to have a ton of JS bloat due to google tag manager, so any KBs we can save for those site are important. 👍🏻

basham:
I assumed that it would be a better workflow having Eleventy as middleware in Vite, but yes, perhaps it is better the other way around. I'll explore that option in another PR.

Yes, I can explore minifying in another PR. Even if it means you can't do tree shaking — it really isn't important for the small amount of things that's in the JS library right now.

PR #796, opened 2025-09-17
Format with Prettier

basham:
Notes:
1. There are no official or community [Prettier plugins](https://prettier.io/docs/plugins) for Nunjucks (but there is a [plugin for Astro](https://github.com/withastro/prettier-plugin-astro)).
2. [Nunjucks](https://github.com/mozilla/nunjucks) is not actively maintained.
3. Liquid is the only other directly supported [templating language for 11ty](https://www.11ty.dev/docs/languages/) other than Nunjucks. Liquid does not have an equivalent feature like macros, which makes it not feasible.
4. In short, if wanting to use a well maintained and well supported environment for the sandbox, I really think [Astro](https://astro.build/) is the way to go.
5. I removed [StandardJS](https://standardjs.com/), as it conflicts with Prettier's defaults.
6. Prettier has solutions for [working alongside linters](https://prettier.io/docs/integrating-with-linters) if we would choose to do that.
7. Updated dependencies.
8. Prettier formats source code. It does not format the output of Vite's `dist` folder.

Future work:
1. Format and add software license banner as a [pre-commit hook](https://prettier.io/docs/precommit). Follow [Rivet React's approach](https://github.com/indiana-university/rivet-react/blob/master/package.json).
2. Replace 11ty with Astro.
3. Explore adding ESLint for code quality checking.


basham:
@levimcg Given this PR includes thousands of lines of newly formatted files, I would recommend you looking at the first three commits on this PR to see how Prettier itself was configured.

levimcg (approved):
Everything looks good to me. I goofed up some of the code and ran the new format task and it fixed everything as expected. Nice!

PR #800, opened 2025-09-22
Refine button component

basham:
Notes:
1. The CTA type is now opt-in (`rvt-button--cta`).
2. The Default type is now named "Form" buttons (`rvt-button`).
3. The disabled style is applied to CTA links without `href`, as these are functionally broken.

Changes:
1. Renamed primary/solid/danger buttons to "strong" (`rvt-button--strong`).
2. Renamed secondary/outline buttons to "normal" (no modifier class).
5. Renamed plain/text buttons to "subtle" (`rvt-button--subtle`).
6. Removed `rvt-button-group--right` in favor of just using `rvt-justify-end`.
7. Replaced `rvt-cta` with `rvt-button rvt-button--cta rvt-button--subtle`.
8. Replaced `rvt-cta rvt-cta--button` with `rvt-button rvt-button--cta`.
9. Replaced `rvt-button--tag` with `rvt-button--small` (24px tall).
10. Added `rvt-button--medium` (32px tall).

Future work:
1. There are a lot of references in the sandbox HTML to CTA and buttons. I did many bulk replacements. I'm sure there are many places that weren't done right and needs to be tweaked. I figure that will be cleaned up as we are working on those particular components/layouts.
2. Buttons with `aria-disabled="true"` can be focused. In this case, the focus color also is reduced in opacity, which is not ideal. This needs further investigation.
3. The pressed style likely needs to be refined.
4. Icons in small buttons are scaled down 16px to 12px square to better fit in that confined space. Maybe we create new icons versions of the more common icons to be cleaner at this smaller size?
5. Segmented button group variant
6. Button with dropdown variant


levimcg (commented):
@basham Awesome, this looks great.

One question just to make sure. With an `<a>` tag if I do the following I get the default black button styles. 

```html
<a href="#" class="rvt-button">Button link</a>
```

image


Which is totally fine with me, I just want to make sure that's the intent.

basham:
Yes, that's the intent. I decided to isolate all CTA styles into `rvt-button--cta`, including the crimson colors and arrows. That will make it easier if we decide to ever introduce other major types beyond the default Form buttons and the CTA buttons.

levimcg (approved):
Looks great 💯

PR #801, opened 2025-09-23
Refine Hero component

basham:
Changes:
1. Removed `.rvt-hero__inner`.
2. Replaced `.rvt-hero__quick-links` with `.rvt-hero__list-hub`.
3. Replaced `.rvt-hero__media` with `.rvt-hero__figure`.
4. Added `.rvt-hero__images` as a child of `.rvt-hero__figure`.
5. Replaced `.rvt-hero__profile-image` with `.rvt-hero__avatar`.
6. Updated `container` mixin to use fluid padding. Using `box-sizing: content-box` to make sure the container contents matches the breakpoint size, which makes things visually more stable.
7. Moved font dependencies from `<link>` to CSS `@import()`.
8. Fix a typo in which some variables were written as `-var()` instead of `var()`.

Future work:
1. Review container and layout styles.


levimcg (approved):
This looks good to me. The new typography feels a bit more unified. Nice improvement.

**Adding a note to myself here re: some of the CSS class name changes**:

The new `.rvt-hero__images` element that is a (child of the newly renamed `.rvt-hero__figure` element) needs updated in the feature, statement, and split hero variants.

PR #804, opened 2025-09-25
Export default

basham:
This makes it so you can use both of these syntaxes:

```js
// Option 1 (current)
import { init } from "rivet.js";
init();

// Option 2 (new)
import Rivet from "rivet.js";
Rivet.init();
```

Future work:
1. Test tree shaking. I suspect tree shaking won't work because the root file imports all component modules for `init()`.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -63,3 +63,16 @@ export {
 	Tabs,
 	init,
 };
+
+export default {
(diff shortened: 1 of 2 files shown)


levimcg (approved):
Awesome, thank you! 🙌🏻

PR #807, opened 2025-09-26
Fix card

basham:
Changes:
1. Renamed `rvt-card-list` to `rvt-card-grid`.
2. Make cards full height in `rvt-card-grid`.
3. `rvt-card-grid` can be applied to a `<ul>`, `<li>`, or `<div>`.
4. Removed `margin-top` on `rvt-card` and `rvt-card-grid`.


levimcg (approved):
Did some testing and poking around and this all looks great to me!

PR #820, opened 2025-10-10
Refine typography

basham:
Notes:
1. Consolidated major typography details into [`font` shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/font) (font family, size, line height, width, weight).
2. Replace font size breakpoints with fluid typography.

Changes:
1. Removed the need for `rvt-items-start` on `<li class="rvt-contact-info__icon-list-item">`.
2. Added `.rvt-avatar__text` to wrap initials.
3. Simplified `rvt-footer-resources` styles. Removed `.rvt-grid`, `__heading`, `__text-block`, `__list`, `__list-item`. Added `__content`, `__address`, `__links`.
4. Added `.rvt-quote--compact`.
5. Replaced `.rvt-stat-group__inner` with `<ul> / <li> / <div class="rvt-stat">`. 
6. Removed `.rvt-stat__content` and `.rvt-stat__image`.
7. Removed `<a class="rvt-stat" href>` styling.
8. Removed `.rvt-prose` (keeping `.rvt-rich-text`).
9. Removed `--rvt-rich-text-stack-space`.
10. Made the default `.rvt-section-intro` left aligned.
11. Added `.rvt-section-intro--center` for center alignment.
12. Removed `.rvt-section-intro--teaser`.
13. Added `size` tokens, which correspond to pixel-equivalent rem values. This is the new base dimension for font size and line height.
14. Added fluid size variables in the format of `--rvt-size-16-18` (which scales 16px to 18px, in rem). The scale is based on viewport size, not container size.
15. Made all text use "pretty wrap" by default.
16. Renamed "regular" font weight to "normal" font weight.
17. Removed "medium" font weight. Using "normal", "bold", "black", and "extra-black".
18. Removed tokens `font-family-base`, `font-family-condensed`.
19. Added `font-width` tokens, which correspond to the [font stretch keywords](https://developer.mozilla.org/en-US/docs/Web/CSS/font-stretch#keyword_to_numeric_mapping).
20. Added `letter-spacing` tokens. The `letter-spacing-wide` is only used for small uppercase text. Unfortunately, letter spacing and text transform is not part of the font shorthand.
21. Added `line-length` tokens.
22. Removed `line-height` tokens in favor of the font shorthand.
23. Expose a network URL in development mode (`vite --host`).

Future work:
1. Clean up utilities.
2. Use the `size` tokens for spacing and border width.
3. Remove `ts` tokens.
4. Generate fluid values through Style Dictionary transforms — so the preferred value doesn't need to be manually created with the [Utopia calculator](https://utopia.fyi/clamp/calculator?a=320,864,18—20&p=rvt-font-size).
5. Make font shorthand a new token type.
6. Remove `link-hub` component.


src/sass/footer-system/_footer-resources.scss
```diff
@@ -1,74 +1,67 @@
...
-		padding-bottom: var(--rvt-spacing-lg);
-		padding-top: var(--rvt-spacing-lg);
+	.rvt-footer-resources__links {
+		display: flex;
+		flex-wrap: wrap;
+		gap: var(--rvt-spacing-md);
 
-		&__heading {
-			margin-top: 0;
+		> * {
```
levimcg, line 49: These columns of links should be equal width like a normal grid. Right now the width is being determined by the flex algorithm based on the size of the content inside leading to uneven column widths.
image

Usually adding something like `> * { flex-basis: 0; }` along with the `flex-grow: 1` you've already set will force all flex children to be equal width, but I'll leave the solution to you.

levimcg (changes requested):
This is looking good. Lot's of nice updates/simplifications 👍🏻

I left one inline comment here re: the footer resources module links columns. Once that's fixed, this all looks good to me.

Making a couple of notes to myself here for when we need to go back and update things in the SB/the CMS.

- This will require significant markup changes to the footer resource module in the core site builder codebase.
- Update avatar markup in quote and profile page type
- Significant markup updates to the stat component
- Remove `rvt-section-intro--teaser` where appropriate. The old "teaser" variant is the new default variant.

src/sass/footer-system/_footer-resources.scss
basham: I added `flex-basis: 15ch`. That seems to be a good balance for the content — so it breaks at a decent place and goes into a single column on mobile.

levimcg (approved):
Looks great, thanks!

PR #823, opened 2025-10-21
Fix hero statement card position in Firefox/Safari

basham:
Notes:
- Resolves [EEDS-112](https://iu-uits.atlassian.net/browse/EEDS-112)

[EEDS-112]: https://iu-uits.atlassian.net/browse/EEDS-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/sass/hero-area/_base.scss
+++ b/src/sass/hero-area/_base.scss
@@ -223,17 +223,17 @@
 }
 
 .rvt-hero--statement {
+	.rvt-hero__body > * {
+		margin-inline: auto;
+		text-align: center;
+	}
+
 	.rvt-hero__title {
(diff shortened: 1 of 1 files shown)


levimcg (changes requested):
This is looking much better. The only thing I'm still seeing is that the hero body/text and CTAs should be centered. The text is centered but the element that wraps it is left-aligned, but should also be centered. See screenshot.

Screenshot 2025-10-21 at 11 16 57 AM

levimcg (approved):
Great, that fixed it!

PR #824, opened 2025-10-21
Reorg files

basham:
Changes:
1. Moved code from `src/sass` and `src/js` to `src/base`, `src/components`, and `src/utilities`.
2. Organized styles by cascade layers: `rivet.base`, `rivet.components`, and `rivet.utilities`.
3. Removed `!important` flags, since cascade layers sets the specificity.
4. Added `rvt-icon` and `rvt-sticker` styles, so they no longer need to be externally included. Due to cascade layers, there wasn't any longer a good way to update the sticker colors short of doing this change — which we wanted to do anyway.
5. Converted some SCSS files to CSS, where the line changes were minimal.
6. Split CSS files that included multiple components.
7. Removed `rvt-flow` on `rvt-card__content` macro — the effect of which was evident and no longer desired now that cascade layers is in place.
8. Removed `new-component` and `sandbox-new-component` npm tasks.
9. Updated dependencies.

Future work:
1. Automatically apply and keep license banners up to date. Should the copyright year reflect the year a file was created or the latest year the project was modified?
2. Declare additional cascade layers around the current layers as hooks for custom code. Maybe `rivet.pre-base`, `rivet.post-base`, `rivet.pre-components`, `rivet.post-components`, `rivet.pre-utilities`, and `rivet.post-utilities`?
3. Migrate dev environment templates to this new folder structure.


levimcg (approved):
This all is looking good to me!

> Should the copyright year reflect the year a file was created or the latest year the project was modified?

It's was my understanding that the copyright year stays is set when the file is created and then doesn't change after that. It is always the year the file was first released, not modified. But, now it looks like individual file headers might not even be required?

From the opensource.iu.edu site:
image

PR #825, opened 2025-10-22
Set up Astro sandbox environment

basham:
Changes:
1. Replaced 11ty with Astro.
2. Consolidated all sandbox content into their respective `src/components`, `src/layouts`, and `src/utilities` folders.

Future work:
1. Reduce redundant content (especially in layouts) with better use of Astro components.


levimcg (approved):
This looks like a great start getting Astro set up!

PR #826, opened 2025-10-28
Use size tokens

basham:
Changes:
1. `size` tokens are now what other dimensional tokens are based on.
2. Replaced many uses of `calc()`, `px`, and `rem` with a size token.
3. There are enough uses of `-1px` that there is now a dedicated token. To avoid it being called `size--1` (which could be confusing, I'm calling it `size-n1` ("n" is for "negative").

Future work:
1. More clean up will happen as we continue refining the components. There are examples of sub-pixel calculations and such that does not yet have a quick fix.

Questions:
1. Should "n1" be "negative-1", "neg-1", "minus-1", "m1" — or just keep it as "n1"?


levimcg (approved):
Nice, this look great. re: the negative margin naming conventions. I think the `n*` format works fine. Spelling out `negative` is the clearest option, but that does make it a bit long.

PR #828, opened 2025-10-30
Work on footer

basham:
Changes:
- The footer is restructured into four components: `rvt-footer`, `rvt-footer-resources`, `rvt-footer-social`, `rvt-footer-legal`
- The HTML is quite different, with `<footer class="rvt-footer">` now being the top-level container. There are no complementary landmarks. The accessibility team is okay with this new direction. "We talked more about the idea of placing the belt and skirt in the footer. That would be fine. The belt and skirt sections have sr-only headings, which makes it easy to navigate to those areas even if they are in a single footer landmark."
- Also: It is best to keep the copyright information last in the footer (and therefore last in the page). This helps with screen reader users to confidently know when they've reached the end of the page of an IU site.


levimcg (approved):
Everything looks good and inline with what we discussed in our face-to-face meeting last week 👍🏻

PR #831, opened 2025-11-13
Reduce uppercase in headings

basham:
Changes:
- Removed most uses of uppercase in headings.
- Tweaked ascender height (YTAS) and lowercase height (YTLC) for type with black font weight (primarily used in headings).
- Adjusted font size (primarily in headings).


levimcg (approved):
Looks good.

PR #833, opened 2025-11-20
Set up config for core and extra icons

basham:
Changes:
1. All files that Style Dictionary generates now includes a `.tmp` postfix, such as `tokens.tmp.css`.
2. Created new "graphic" token type, which includes the values for path, width, height, container width, and container height. This is used for any graphic that isn't a standard dimension, such as icons and stickers. That is currently limited to the IU logo and the open quote.
3. Created new `<rvt-logo-iu tab>` element, which replaces `rvt-lockup__tab` and the inline SVG.
4. Icons are now designated as being part of the "core" `rivet.css` file or part of the optional `rivet-icons.css` file.
5. There is no more need for the external `rivet-icons.js` file.
6. Fixed the size of the social icons to be from 48px square to 16px square, matching the other icons. These still get scaled down to 12px to fit how we use them in the footer.
7. Removed the unused `tokens/namespace.json` file.
8. Replaced all size token values with the pixel values, since now Style Dictionary will transform it to the appropriate `rem` value.
9. Updated dependencies.

Notes:
1. I suspect there's a bug in `rimraf`. In order to delete all `tmp` files that Style Dictionary generates (see the npm script `build:clean`), I have to add arguments for each folder level that they reside in — instead of it working recursively. Odd.

Future work:
1. Add more sizes to the `rvt-logo-iu` element as needed.
2. Confirm with the Branding team about the redrawn IU logo, the dimensions of the tab, and about updating the favicon.


levimcg (approved):
Nice, this looks good as far as I can tell.

PR #835, opened 2025-12-12
Create Text Essay component

basham:
Changes:
1. Added `rvt-text-essay` component. By default, it displays content in a single column with centered headings. Use variant `rvt-text-essay--wide` to enable a two column layout with left-aligned headings on wider viewports.
2. Moved `rvt-rich-text` from utilities to components. This solves a specificity issue with the Text Essay component. Generally, a class that uses element selectors should be considered a "component." A class that specifically modifies only its applied element and not descendants should be considered a "utility."


levimcg (approved):
Looks great!

PR #836, opened 2026-01-06
Add stickers

basham:
Changes:
1. Updated dependencies.
2. Opt-in to stickers by linking to: `dist/rivet-stickers.css`
3. These external dependencies are longer needed: `rivet-sticker-element.js` and `rivet-stickers.js`
4. Stickers are now defined in `src/tokens/sticker.json`. Each sticker token is a composite of two values: the white "path-fill" and the colored "path-stroke".

Renamed stickers:
- badge-check → shield-check
- badge-lock → shield-lock
- check-pen → pen-check
- notes-exclamation → page-exclamation

New stickers:
- ambulance
- beaker
- bone
- brain
- droplet
- envelope-bell
- head-brain
- lungs
- molecular-structure
- music-note
- page-code
- petri-dish
- ribbon
- stethoscope
- violin
- virus
- wheelchair

Future work:
- `rivet-stickers.css` is 180kb, which is much more than `rivet-icons.css` (52kb) and nearly as much as `rivet.css` (240kb). Explore if `rivet-stickers.css` needs to be broken down into submodules. It may not be worth it, though.


basham:
Review the new sticker pages: http://localhost:4321/components/sticker

levimcg (approved):
This looks great.

RE: file sizes CSS vs. old JS lib. I checked the compressed/gzip size of of rivet-stickers.css with a command line tool here is the breakdown:

- `rivet-stickers.css` (gzip) - 49K
- `rivet-stickers.js` (gzip) - 38.1K

All our servers/containers in the web environment have gzip enabled by default. So, `rivet-stickers.css` is about 10K bigger compressed which seems like a decent trade-off to me if it eliminates another JS library to load an parse.

basham:
And also with that comparison, their size would be even closer if `rivet-stickers.js` had the same number of stickers as `rivet-stickers.css`.

basham:
The extra size is likely due to the `name` selector, which is repeated for each sticker. GZIP certainly saves a lot of space on that repetition, though.

```css
rvt-sticker[name="accessibility"] {
	--path-fill: var(--rvt-sticker-accessibility-path-fill);
	--path-stroke: var(--rvt-sticker-accessibility-path-stroke);
}
```

PR #850, opened 2026-04-06
Add font GRAD axis

basham:
Changes:
1. Updated Google Fonts URL to add the GRAD ("grade") axis.
2. Added a new variable `--rvt-font-variation-grade`.
3. This makes it so content can be more bold without affecting the text width. This will support the work needed for navigation and styles related to `aria-current="page"`.

Future work:
1. Include the full font file for all Rivet fonts in the `dist` folder. That will just give us more flexibility to use the fonts as we want and not be dependent on Google servers.
2. Consider new naming convention for `--rvt-font-variation-*` variables. `black` is named  "black" because it is meant to be paired with the "black" fonts. In contrast, "grade" isn't named to be paired in a way like that.

### DIFF
--- a/src/base/fonts.css
+++ b/src/base/fonts.css
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-@import url("https://fonts.googleapis.com/css2?family=Azeret+Mono:ital,wght@0,100..900;1,100..900&family=Playfair+Display:ital,wght@0,400..900;1,400..900&family=Roboto+Flex:opsz,slnt,wdth,wght,YTAS,YTLC@8..144,-10..0,25..151,100..1000,649..854,416..570&display=swap");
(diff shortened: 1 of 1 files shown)


basham:
Another reason to not use `@import` for fonts:

The content order in `rivet.css` is the licensing comment, then `@charset "UTF-8"`, then the `@import` font by URL.

I'm noticing a warning when importing `rivet.css` from a separate Astro project:

```
13:19:05 [WARN] [vite] [vite:css][postcss] @charset must precede all other statements
5  |   * SPDX-License-Identifier: BSD-3-Clause
6  |   */
7  |  @charset "UTF-8";@import"https://fonts.googleapis.com/css2?family=Azeret+Mono:ital,wght@0,100..900;1,100..900&family=...
   |  ^^^^^^^^^^^^^^^^^
8  |  
```

We can either somehow move the licensing comment or if we instead reference a local (not remote) font file, then we may not need `@charset`. This all needs some exploration.

levimcg (approved):

PR #852, opened 2026-04-06
Add Astro action for deploying to GitHub Pages

basham:
Changes:
1. This action will allow automatically deploy the Astro sandbox in the `v3-main` branch to GitHub Pages on every push.

Note: We'll need to make a few quick changes to the repo's settings to get this to work after it merges.


levimcg (approved):
Nice, let me know if there's anything I can do to help re: repo settings once this merges, but I think you should have admin access to update most settings.

PR #854, opened 2026-04-07
Fix sandbox environment for GitHub Pages

basham:
Changes:
1. Astro sandbox gets built during `build` script and outputs to `dist-astro` folder (which GitHub Actions then deploys).
2. Links are now relative to the base URL ("/rivet-source").
3. Added the package version number to the sandbox header and page title.

Future work:
1. The sandbox watches and compiles the Rivet source code on its own. Instead, it should use the Rivet output in the `dist` folder.


levimcg (approved):

basham:
The deploy worked!

https://indiana-university.github.io/rivet-source/

basham:
Although, I forgot to add a note in the Readme about it. I'll work on that in another PR.

PR #859, opened 2026-04-27
EEDS-210: Fix billboard reverse

basham:
Notes:
1. See: [EEDS-210](https://iu-uits.atlassian.net/browse/EEDS-210)
2. I added a new sandbox page to show the reverse variant.

[EEDS-210]: https://iu-uits.atlassian.net/browse/EEDS-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/billboard/billboard-reverse.astro
+++ b/src/components/billboard/billboard-reverse.astro
@@ -1,0 +1,7 @@
+---
+import Billboard from "./billboard.astro";
+
+export const title = "Billboard Reverse";
+---
+
+<Billboard reverse />
--- a/src/components/billboard/billboard.astro
+++ b/src/components/billboard/billboard.astro
@@ -1,8 +1,14 @@
 ---
 export const title = "Billboard";
+
+const { reverse = false } = Astro.props;
+const classList = [
+	"rvt-billboard",
+	{ "rvt-billboard--reverse": reverse }
+];
 ---
 
-<div class="rvt-billboard">
+<div class:list={classList}>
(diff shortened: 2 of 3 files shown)


levimcg (commented):
The padding fix on the reversed variant looks good. I did notice this while testing on smaller screens. Is there a need for this max-width of 25rem on the billboard image instead of just leaving it fluid to fill up the space of the parent once it changes from a horizontal/column to a vertical/row orientation?

This could be addressed separately if you want, but unless there is a specific purpose for that max-width, I would expect the image to fill up the space of it's parent element.

Image

basham:
Rationale for `max-width`:
1. The billboard image is meant to always be a square aspect ratio.
2. If the image grows to the full container size at the large end of the single column layout, it pushes the content really down the page.
3. The `max-width` makes a smoother transition at the point that it goes from one column to a two column layout, because the image size is the same dimension at this breakpoint. Without it, the image would be really big at one column, then small again when it is two columns; it would make the transition really dramatic.
4. I would like a better solution, but I'm not sure what it is. Would be open to discuss other options with you, Jenn, and I.

levimcg:
I see, that makes sense to me. I generally get questions like this from marketing design/content folks and other building site in Site Builder. Just wanted to make sure I could explain the intent.

levimcg (approved):

PR #879, opened 2026-06-03
Refactor into monorepo with pnpm

basham:
## Overview
- This PR refactors the entire repo into a monorepo with [pnpm](https://pnpm.io/).
- Packages are under the `packages` folder. Currently: `tokens` (public), `core` (public), and `sandbox` (private).
- Packages output to their own `dist` folder, which can be referenced by other packages.
- The root package just handles scripts that affect the entire repo, like linting.
- Because pnpm handles dependencies better and because the sandbox and core packages are separate, they are each able to run on different versions of Vite. Core runs Vite v8, while Astro still depends on Vite v7.
- I removed the unused generate-changelog task/script.
- `rivet.js` now automatically initializes.

## Workspace dependences

If a package depends on another package in the same monorepo, then declare it as so in its `package.json`. For example, `@rivet-iu/core` depends on `@rivet-iu/tokens`.

```json
"dependencies": {
	"@rivet-iu/tokens": "workspace:*"
}
```

Then import it where needed, in either CSS or JS. For example:

```css
/* rivet-icons.css */
@import "layers.css";
@import "@rivet-iu/tokens/dist/tokens-icon-extra.css" layer(rivet.base);
@import "@rivet-iu/tokens/dist/icon-extra.css" layer(rivet.components);
```

## Testing

Install [pnpm](https://pnpm.io/). I recommend [installing it with Homebrew](https://pnpm.io/installation#using-homebrew).

When running a pnpm command, it will automatically download, install, and uninstall dependencies if/as needed to align with the `package.json` files. There's no need for an explicit `install` command.

Start the sandbox dev environment (with hot module reloading).

```
pnpm dev
```

Preview the sandbox environment before it deploys to GitHub Pages (without hot module reloading). It copies  (no longer modifies) the `core/dist` folder, so it now more fully mimics a production environment.

```
pnpm preview
```

Build the packages to see their `dist` output.

```
pnpm build
```

## Access the published package

I've removed the generation of the `rivet.zip` file. That was used as a convenient way to copy and send the `dist` files around. Instead, I recommend one of the following:

View the package contents on the npm package page:
https://www.npmjs.com/package/rivet-core?activeTab=code

Download the original `.tgz` file, following this pattern:
https://registry.npmjs.org/rivet-core/-/rivet-core-3.0.0-alpha.10.tgz

## Changes to `package.json`

1. There is now the root `package.json` and ones in each package. So, four total.
2. The order of these field names are normalized across these files. There seems to be just weak community conventions concerning their order, so I tried grouping things *logically*.
3. The [`"exports"` field](https://nodejs.org/api/packages.html#exports) is the modern replacement to `"main"` (CJS) and `"module"` (MJS).
4. Unfortunately, the only thing that [Unpkg](https://unpkg.com/) and [jsDelivr](https://www.jsdelivr.com/documentation#id-configuring-a-default-file-in-packagejson) agrees on for the default export is to use the `"main"` field. I'd prefer to get rid of the `"main"` field and just rely on the `.` as the default export but that doesn't seem realistic right now if we want to keep the shortcut. However, given we have so many relevant files is our package `dist` folders, I don't think having a "default export" really makes too much sense. The default export for `core` is `rivet.js` but the most important file is actually `rivet.css`. For `tokens`, maybe `tokens.json` is the most relevant? We have such a mix of things, it just seems odd having any main or default. I'd like to advocate to just force users to point directly to the `dist/rivet.js` file. It results in less "magic" and clarifies exactly what they are wanting.
5. From what I understand, the `"publishConfig"` field is required for publishing public packages to a scope (`@rivet-iu`).
6. I'm including `"version"`, `"homepage"`, `"repository"`, and `"bugs"` for all public packages but not for private packages.
7. I've added the [`"imports"` field](https://nodejs.org/api/packages.html#subpath-imports) which provides convenient shorthands when importing code from the same package.
8. I've exported `package.json`, so that the version number (and anything else in there) can be looked up. It is automatically included in the pack, so there isn't a need to add it to the `"files"` field.


levimcg (approved):
This looks great as far as I can tell, really well thought out! It was really straightforward to install pnpm with Homebrew, all of dev, preview, and build commands worked great. I'm sure I'll needs a few more walkthroughs of how things work, but this feels like a really nice improvement.

I think it might be worth getting together for 30 minutes sometime soon to talk about what publishing a next release looks like, but it seems pretty straightforward from what I can tell in the pnpm documentation. https://pnpm.io/cli/publish

PR #897, opened 2026-08-06
Clean up header

basham:
## Notes
- [EEDS-134](https://iu-uits.atlassian.net/browse/EEDS-134)
- This should not involve any HTML changes.

## Changes
1. Dropdown menu will no longer move focus to the first interactive item if it opened due to a mouse event.
2. Priority Nav now uses `<rvt-icon>` for the "More" chevron rather than using an inline SVG.
3. Left aligned menu item labels to align with the dropdown menu label.
4. Dropdown uses anchor positioning API to ensure the popover menu does not position outside of the viewport.
5. Improved contrast in all color themes.
6. Menu items no longer span multiple lines.
7. Menu width is dynamic to the menu item content.
8. Aligned the `aria-current` style with other v3 work.

## Polyfill

[Anchor positioning is Baseline Newly Available](https://caniuse.com/css-anchor-positioning), having just made that threshold in January 2026.

I added the [CSS Anchor Positioning Polyfill](https://anchor-positioning.oddbird.net/), but I ended up not using it, as I wasn't able to get it working accurately. I'd like to leave it in the code for now, as perhaps we can revisit it later on. To enable the polyfill, just include it in `rivet.js`:

```js
import "./polyfill.js";
```

Once imported, the polyfill will be included in the `dist/polyfills` folder when running `pnpm build`.

Instead of the polyfill, I just used `@supports` to use absolute positioning if anchor positioning is not available. If the menu item is the last one, then that menu is right aligned. There could be some use cases in which this breaks, but this is still an improvement over the current implementation.

The only way I could test the lack of this feature is by using Firefox:

1. Go to `about:config`.
2. Search for `layout.css.anchor-positioning.enabled`.
3. Toggle the value to `false`.
4. Refresh the page.

> [!NOTE]
> The Rules panel in the Firefox dev tools is empty and does not work when this config is set. I bet it would be the same for others as well, but I didn't test that.

## Future work
1. Clean up HTML and CSS classes.

[EEDS-134]: https://iu-uits.atlassian.net/browse/EEDS-134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


levimcg (approved):
This all looks great to me.

Adding a note here more for myself that this will require a minor update to the HTML in the Site Builder implementation of the header. It currently uses inline SVG for the `chevron-down` icon in the navigation dropdowns, but that's a pretty straightforward swap.

